PR #14, opened 2018-09-02
first version of prioritized dqn

jbrea:
TODO
* [ ] write simple tests
* [ ] test with Atari
* [ ] simplify code


findmyway:
Now we have the `SumTree` here: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/master/src/Utils/sum_tree.jl
Let's use https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/27 to track the progress.

PR #17, opened 2018-09-07
[WIP] Refactor Buffer related code

findmyway:
There's still a lot of things to do.

- [x] Replace old Buffers
- [ ] Pass Tests


findmyway:
Hi @jbrea 

Now I fixed this [TODO](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/17/files#diff-22e0b46fe3cb125d7061a732de0f1f35L37) and make the code more concise here. More simplifications are coming soon.

In the old buffer implementation, the length of actions and states is different (with 1-offset). Now I force the states/actions/rewards/done have the same length to avoid potential errors.

Most of the tests are passed, however, there are still two test errors. I'm not very sure why these two test cases fail. Can you help to check whether I missed something or if the test case needs to be updated?

1. https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/17/files#diff-1df9849708fd0eb8be63e29b738bcb58R17
2. https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/17/files#diff-fe5ee5a34f9caa47222aa71775b80eafR10

Thanks in advance.

zsunberg:
Hi @findmyway and @jbrea ,

I've been casually watching the development of this package, and noticed this might deal with some issues that we also dealt with in the POMDPs.jl framework (which also includes a lot of MDPs). We have found the best way to talk about "steps" or "stages" or "turns" in an MDP is with (s,a,r,s') tuples instead of just (s, a, r) as it appears that you're trying to do here. This gets rid of the ambiguity about whether a state/observation/reward is the *result* of an action or the previous context that an action is taken in. Just thought that might be relevant given that you're dealing with state/action vectors that are different lengths and `buffer.states[end]` being the "next turn's state".

If more details about our experience would be relevant, let me know (the above is not a very thorough/clear explanation, but didn't want to waste too much time if it's not relevant).

findmyway:
Hi @zsunberg 

Thanks for sharing your precious experience!

I think your suggestion is really great. 
Could you also kindly share the implementation detail?

Currently we have a buffer struct like this:

```
 struct Buffer{Ts, Ta}}
     states::CircularBuffer{Ts}
     actions::CircularBuffer{Ta}
     rewards::CircularBuffer{Float64}
     done::CircularBuffer{Bool}
 end
```

Do you mean that we'd better add another field to store the following-up states?

<hr>

At the first glance, it seems a little waste to store the the following-up states.
But after a second thought, I think this solution is considerable.

1. Like Zachary said, it will reduce many ambiguities and some corner cases (like the end of an episode)
2. This kind of buffer will be more general to handle async environments. Because in that case, the next turn's `s` may not be the same with current turn's `s'`, and our assumption is not valid anymore.

What do you think? @jbrea

zsunberg:
So we use the MDPHistory and POMDPHistory types that are implemented here: https://github.com/JuliaPOMDP/POMDPSimulators.jl/blob/master/src/history.jl

They are accessed through the `eachstep` iterator.

Say you have a history object `hist`, and you wanted to print the start state, action, and end state of each transition, then you could use
```julia
for s, a, sp in eachstep(hist, "s,a,sp")
    @show s
    @show a
    @show sp
end
```
Or if you just wanted a vector of the rewards, you could do
```julia
collect(eachstep(hist, "r"))
```

zsunberg:
That seems to be flexible enough for people to get what they want from the history, especially in the POMDP case where there are more things: states, observations, beliefs, etc. it also has support for the "info" that is output by the model or solver

zsunberg:
`eachstep` kind of hides the implementation details from the user so things are less ambiguous

findmyway:
Great thanks for providing the `history.jl` file!
I'll take a close look and borrow some ideas here.

jbrea:
Hi @zsunberg 

Thanks for sharing your insights! The eachstep function is indeed nice for the user.

@findmyway 
> Do you mean that we'd better add another field to store the following-up states?

If I understand correctly, @zsunberg just accesses [the next element in the history vector](https://github.com/JuliaPOMDP/POMDPSimulators.jl/blob/94a91ae9338b8d02a724a49451e2aa9291cb5c96/src/history.jl#L218) to show the next state. This is convenient for the user but doesn't use extra memory.

> In the old buffer implementation, the length of actions and states is different (with 1-offset). Now I force the states/actions/rewards/done have the same length to avoid potential errors.

I found it most natural to have this offset. In the first step we start with observation s0, take action a0 and observe r0, isdone0 and state s1: there are two states, but only one action, reward and end of episode signal. Note that also in @zsunberg's approach there always has to be a ```state[i+1]``` to access the next state, but only ```action[i]```, ```reward[i]``` etc.

> Seems like the reason is that I ```reset!(environment)``` in the first step in ```learn!```.

I preferred not to reset the environment whenever ```learn!``` is called. The reason is basically that I did not like the idea of resetting the environment when I want to continue learning. Say I learn for N steps and end in the middle of an episode: why should the environment be reset when I decide to continue learning for  another N steps?

While the code may not look very elegant, a lot of thought went into designing the buffers and the ```learn!``` function. I would be happy to explain all design choices that seem unclear. And if we find nicer solutions: even be better :)

zsunberg:
An alternative to having a `Turn` structure is using named tuples. That is what we are doing now. Then you don't have to commit to including or not including next states and actions.

jbrea:
Hi @findmyway 

Sorry, I may have written too much in favor of my old buffer implementation: there are certainly ugly parts in the code and you made already so many nice contributions that also this time I may underestimate the elegance of your approach. 

I think the reinforcement learning package will be a success, if 1) the code is elegant 2) easy to extend and 3) runs fast. To get some hints about point 3) I started writing [some benchmarks](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/tree/master/test/benchmarks) that also serve as a test, if the different repositories in this org are consistent.

jbrea:
Hi @findmyway 

1. Ah, I think I get you now: Your point is that we should not wait for the ```update!``` to finish before taking the next ```step!``` in the async setting, right? This is a very good point. I have not yet carefully thought about the async case and I think it is important and a good moment to figure this out now. What I had in mind was roughly something like this
```julia
# worker 1 (replace the 1's with i for worker i)
while true
    s, r, d = interact!(env1, a)
    a = policy(s)
    put!(obs_channel[1], (s, a, r, d))
    policy = take!(pol_channel)
end

# main worker / learner
# typeof(buffers) = Array{Buffer, 1}
while true
    for i in 1:nworkers push!(buffers[i], take!(obs_channel[i])) end
    update!(learner, buffers)
    put!(pol_channel, policy)
end
```
This would require to write a new ```learn!``` function for the async case and also new ```update!``` functions that take arrays of ordinary buffers as input. Do you think this is a good idea? What kind of design did you have in mind?

2. Uups, yes, in the corner case the last state is thrown away. I think there is a ```if fillbuffer; pushstate!(buffer, s) end``` missing inside the ```if done``` block. However, I think the last state is never needed, since there is no action based on the last observation of an episode; in fact, I believe ViZDoom does not even return an observation at the end of an episode. But throwing away an observation probably screws the circular buffer! Thanks for looking at this so carefully!

findmyway:
1. I don't have a good idea yet. I think your proposal is just great to me.
2. So if it is not a big issue, I'll make some modifications soon :laughing:

jbrea:
Argh, no, I got confused (which means at least that this should be better documented). There is no action and no return based on the terminal state of an episode, which is why it should be thrown away. The circular buffer doesn't get screwed. 

My thinking for the original design was the following:
1. When ```learn!``` is called the very first time, we don't know the state of the environment, so let's get the state and select the first action ```firststateaction!(rlsetup)```. This pushes one state-action pair to the buffer.
2. In each ```step!``` we interact with the environment, push the observed return and isdone to the buffer, observe the new state (which is an initial state if isdone), select the next action and push this state-action pair to the buffer.
3. If we resume ```learn!``` there is already the action for the last state in the buffer, which is why ```firststateaction!(rlsetup)``` returns ```buffer.actions[end]```.

Since in step 1 we push a state-action pair and in each step 2 we push (reward, isdone, state, action), the circular buffer for states and actions needs to be longer by one entry than the reward and the isdone buffer.

An almost equivalent, seemingly simpler alternative would be
1. When ```learn!``` is called the very first time, we don't know the state of the environment, so let's get the state ```firststate!(rlsetup)```. This pushes one state to the buffer.
2. In each ```step!``` we select an action and push it to the actions buffer, interact with the environment, push the observed return and isdone to the buffer, observe the new state (which is an initial state if isdone), and push this state to the buffer.
3. If we resume ```learn!``` there is already the last state in the buffer, which is why ```firststate!(rlsetup)``` returns ```buffer.states[end]```.

Since in step 1 we push a state and in each step 2 we push (action, reward, isdone, state), the circular buffer for states needs to be longer by one entry than the action, reward and the isdone buffer.

Because in SARSA we also need the next action, I went for the first version.

PR #18, opened 2018-09-07
fix examples

jbrea:


jbrea:
@findmyway Can I merge this PR? Otherwise the examples don't work.

findmyway:
Hi @jbrea 

Can we just wait a moment?
Because I made a lot of change about the `Buffer` related code.
I'll explain it in another thread.

jbrea:
I'll merge this now, such that we can tag a version with working examples.
There will only be few merge conflicts with #21.

PR #21, opened 2018-09-14
[WIP] Refactor buffer

findmyway:
Sorry that I closed the [old PR](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/17) by accident.

Will use this one to track the progress.


src/learn.jl
```diff
@@ -1,47 +1,36 @@
-@inline function step!(rlsetup, a)
-    @unpack learner, policy, buffer, preprocessor, environment, fillbuffer = rlsetup
-    s0, r0, done0 = interact!(environment, a)
-    s, r, done = preprocess(preprocessor, s0, r0, done0)
-    if fillbuffer; pushreturn!(buffer, r, done) end
-    if done
-        s0 = reset!(environment)
```
jbrea: I think it should be ```s0, = reset!(environment)``` because ```reset!``` returns a ```NamedTuple{(:observation,),...```

src/learn.jl
```diff
@@ -1,47 +1,36 @@
...
-    if fillbuffer; pushstate!(buffer, s) end
-    a = policy(s)
-    if fillbuffer pushaction!(buffer, a) end
-    s0, a, r, done
-end
-@inline function firststateaction!(rlsetup)
-    @unpack learner, policy, buffer, preprocessor, environment, fillbuffer = rlsetup
-    if isempty(buffer.actions)
-        sraw, done = getstate(environment)
-        if done; sraw = reset!(environment); end
```
jbrea: See line 7.

test/linfuncapprox.jl
```diff
@@ -72,7 +39,7 @@ function testlinfuncapproxflux()
     reset!(env)
     x2.policy.t = 1
     learn!(x2)
-    @test x.learner.net.W.data ≈ x2.learner.params
+    @test_broken x.learner.net.W.data ≈ x2.learner.params
```
jbrea: With option ```replaysize=1``` for DQN the test passes.

src/learner/dqn.jl
```diff
@@ -55,11 +57,11 @@ function DQN(net; kargs...)
...
 function defaultbuffer(learner::Union{DQN, DeepActorCritic}, env, preprocessor)
     state = preprocessstate(preprocessor, getstate(env)[1])
-    ArrayStateBuffer(capacity = typeof(learner) <: DQN ? learner.replaysize :
-                                                         learner.nsteps + learner.nmarkov, 
-                     arraytype = typeof(state).name.wrapper,
-                     datatype = typeof(state[1]),
-                     elemshape = size(state))
+    action = sample(actionspace(env))
+    CircularTurnBuffer{Turn{typeof(state), typeof(action), Float64, Bool}}(
+        typeof(learner) <: DQN ? learner.replaysize : learner.nsteps + learner.nmarkov,
```
jbrea: Should be ```typeof(learner) <: DQN ? learner.replaysize : learner.nsteps```. This also fixes the second test of linfuncapprox.

jbrea:
Hi @findmyway 

Looks really nice so far. :smile:
I need a little bit more time, however, to look at everything; maybe later today, otherwise later this week.
Did you try to run benchmarks_cpu.jl?

test/linfuncapprox.jl
findmyway: Thanks,  resolved now.

src/learner/dqn.jl
findmyway: resolved not

findmyway:
Thanks for your feedback!

There's no need to hurry. :laughing: 
I'll run the benchmarks file tomorrow(seems we have several hours' time difference :sleeping: ) and fix potential efficiency problems.  And resolve some problems caused by the change of environment's return type

src/learner/dqn.jl
```diff
@@ -55,11 +57,11 @@ function DQN(net; kargs...)
...
 function defaultbuffer(learner::Union{DQN, DeepActorCritic}, env, preprocessor)
     state = preprocessstate(preprocessor, getstate(env)[1])
-    ArrayStateBuffer(capacity = typeof(learner) <: DQN ? learner.replaysize :
-                                                         learner.nsteps + learner.nmarkov, 
-                     arraytype = typeof(state).name.wrapper,
-                     datatype = typeof(state[1]),
-                     elemshape = size(state))
+    action = sample(actionspace(env))
+    CircularTurnBuffer{Turn{typeof(state), typeof(action), Float64, Bool}}(
+        typeof(learner) <: DQN ? learner.replaysize : learner.nsteps,
```
jbrea: Sorry, I think I was wrong here before. I guess it should be
```typeof(learner) <: DQN ? learner.replaysize : learner.nsteps + learner.nmarkov - 1```

**Comment by findmyway** [comment]
I did some benchmark tests. And the result showed about 5~10% slower than your original code.
Too sad :disappointed: 

In fact, the speed of of newly written `CircularArrayBuffer` and `CircularTurnBuffer` are just as fast as the old ones(in some cases even a slightly faster).

But for scalar states, the getindex method of `CircularArrayBuffer` is not as efficient as `CircularBuffer` in `DataStructures`(I haven't figured out why). And in the original implementation, the **DEFAULT** buffer is `Buffer` which uses `CircularBuffer` to represent states/actions/rewards/isdone info. Making it pretty fast in most cases.

However, if state space is a larger Array, the difference is not that obvious. Unfortunately, I get an error while running the last example in the benchmark file. Maybe you can help to take a look.(I guess the reason is the dimension change of `nmarkovgetindex`)

It's really an enjoyable experience to read your code and walk through all your performance considerations. I personally take this as a challenge.

So if the performance decrease is unacceptable. maybe you'd merge https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/18 first and let's move on to implement many more interesting algorithms.
Sorry for wasting your precious time reviewing my code.

**Comment by jbrea** [comment]
Hi @findmyway 

You are doing a fantastic job at refactoring my code.
I think many places look already much nicer than before,
and I feel more confident, if the core is not only written by me,
but improved by you.

Regarding performance: I believe we will get there.
Because performance is something, where I believe a Julia RL package can shine,
I would like to have keep a watchful eye on it. 
But if the performance decrease is 5% for small problems, I don't worry.

To get the atari benchmark to work I made the following hacks for now. This could certainly be done nicer.
- In ```learn.jl``` after line 17, I added ```r = float(r)```, because the Atari domain seems to return a non-float reward.
- In ```policies.jl:135```, I added the last dimension ```p.policy(reshape(st, sz[1:end-2]..., sz[end-1] * sz[end], 1))```
- In ```dqn.jl```, I sample indices not from 1 and reshape a bit differently:
```julia
    indices = StatsBase.sample(learner.nmarkov + 1:length(b) - learner.nsteps + 1, 
                               learner.minibatchsize, 
                               replace = false)

    sd = viewconsecutive(b, :states, indices, learner.nmarkov)
    sz = size(sd)[1:end-1]
    qa = learner.net(convert(Array, reshape(sd, sz[1:end-2]..., sz[end-1] * sz[end], learner.minibatchsize)))

    st = viewconsecutive(b, :nextstates, indices .+ (learner.nsteps - 1), learner.nmarkov)
    qat = learner.targetnet(convert(Array, reshape(st, sz[1:end-2]..., sz[end-1] * sz[end], learner.minibatchsize)))
```
What is the philosophy behind this reshaping?
In standard supervised learning the input images are often provided in the shape ```(width, height, color_channels, minibatch_size)```. In a `n`-step Markov setting, a single input is not a single image, but the last `n` images. A natural way to include the last `n` images is to extend the ```color_channels``` to ```n * color_channels```. For the policy the `minibatch_size` is 1 and during learning it is whatever we set it to.

As soon as I find a bit of time I will also write a short GPU benchmark file. I think the current version with explicit conversion to `Array` would not yet work on the GPU, where rather a `CuArray` or something similar is expected.

PR #22, opened 2018-09-26
Refactor buffers

jbrea:
Hi @findmyway 

I tried my best to continue your effort on refactoring the buffers, until I found an efficient way to write `getindexconsecutive` for `CircularBuffers`. The performance stays the same as before but I think the code looks much nicer now.

I did not yet adapt the rest of the code to the interface functions `states`, `nextstates` etc. at the bottom of the buffers file. But I thought this would be a nice and efficient way to abstract away the specifics of the buffer implementation.

What do you think?

### DIFF
--- a/src/buffers.jl
+++ b/src/buffers.jl
@@ -80,6 +80,110 @@ mutable struct ArrayCircularBuffer{T}
     counter::Int64
     full::Bool
 end
+
+import Base: (==), size, getindex, setindex!, length, push!, empty!, isempty, eltype, getproperty, view
+mutable struct CircularArrayBuffer{A}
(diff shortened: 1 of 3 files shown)


findmyway:
Just come back from vacation :smile: 

I totally agree with the gradual and piecemeal change :+1:

findmyway:
`getindexconsecutive` is renamed to `consecutive_view` here: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/master/src/components/buffers/circular_array_buffer.jl#L155-L158

It is not exactly the same but can reach similar effect.

PR #25, opened 2019-07-30
[WIP]Reorganize Existing Components

findmyway:
Just want to let you guys know that I'm still working on improving this package.

Generally speaking, there are three major steps in my mind:

1. Reorganize the code structure and keep some common components well tested.

    - [x] Move some useful components from [Ju.jl](https://github.com/Ju-jl/Ju.jl/tree/master/src) here.
    - [x] Move preprocessing components to [ReinforcementLearningEnvironments.jl](https://github.com/JuliaReinforcementLearning/ReinforcementLearningEnvironments.jl).

1. Rewrite existing algorithms using a set of unified interfaces.
    - [x] td
    - [ ] mdp
    - [ ] montecarlo
    - [ ] prioritized sweeping
    - [x] q-learning
      - [x] dqn
      - [ ] prioritized experience replay
      - [ ] double dqn
      - [ ] dueling dqn
      - [ ] retrace
      - [ ] noisy dqn
      - [ ] distributed dqn (C51)
    - [x] reinforce
      - [ ] PPO
      - [ ] DPPO
      - [ ] TRPO
    - [ ] actor critic
      - [ ] a3c
      - [ ] ddpg
      - [ ] td3
      - [ ] soft actor critic

    I haven't tried to implement any model-based methods in Julia yet. So this part is left out above. In the early stage, I will only implement several algorithms above to verify that the interfaces are flexible enough. Then I'll move on to the next step.

1. Leverage the multi-threading tasks to allow auto-scheduling on one node and data sharing across different nodes.

    This step should be independent with the steps above. And I have a raw implementation of a global scheduler in my mind. A more practical issue is that I don't have that much machine for some experiments.


jbrea:
Awesome, please let me know if I can contribute to this PR in some way.

findmyway:
image

Hi @jbrea , above is the current concept structure.

Some big changes since the v0.3.0
1. Buffer related data structures are simplified **again** to make a lot of things easier
1. The concept of `Policy` is removed. I personally think this concept is too general and the `Agent` can be a good replacement for now.
1. Multi-stage callbacks are introduced according to @iblis17 's suggestion and they simplify the runtime logging/statistics a lot.
1. I haven't figured out how to move the original MDP related algorithms in the above structure. Still need some more time to think.
1. A lot of components are organized under the above concept structure so that we can provide a set of unified interfaces for both traditional RL and deep learning related RL. 
    - As for distributed RL, a global scheduler may be introduced under the learner level. But I'm really not that sure yet. I've contacted several people in the last several weeks. Some core devs in Julialang seems working on scheduling tasks across threads and remote processors efficiently. But that's a long story. We can't just stop here and wait for new features to come. So I think that for now this package can serve as the following two roles:
        - **Educational**: I'd move [ReinforcementLearningAnIntroduction.jl](https://github.com/Ju-jl/ReinforcementLearningAnIntroduction.jl) under this org if you agree. I find that many people just don't know how to get started with RL in Julia. Those examples can serve as good tutorials.
        - **Experimental**: A lot of mature algorithms (working on one node) will be provided in this project. And I'm quite confident that those built-in components will also be useful to explore some new algorithms.

Some other useful components will also be added:

1. Logging
    - [TensorBoardLogger.jl](https://github.com/PhilipVinc/TensorBoardLogger.jl) works very well according to my experiments.
1. Checkpoint Management
    - So that we can easily reproduce some basic results.

Any suggestions are welcome!

When do you think would be the appropriate time to merge these changes into master?

For now, there's only a simple DQN. And there's still tons of work to do:

1. Rewrite ReinforcementLearningAnIntroduction.jl (~5weeks)
1. Finish DQN related algorithms (~4weeks)
1. Finish policy-based algorithms (~4weeks)
1. Some other algorithms listed above (~4weeks)
1. Built-in TensorBoard logging support (~1week)
1. Check-point management (~1week)
1. Update ReinforcementLearningEnvironments.jl (~1week)

The estimations above are based on my current working hours on this project (about 20 weeks in total). If someone else would like to work on this together, things should happen much faster.

src/components/agents/q_agent.jl
```diff
@@ -0,0 +1,39 @@
...
+    role::String
+    learner::Tl
+    buffer::Tb
+    selector::Ts
+    batch_size::Int
+    update_horizon::Int  # starts with 1
+    γ::Float64
+    act_step::Int
+    min_replay_history::Int
+end
```
jbrea: I always thought of `batch_size`, `update_horizon`, `γ` and `min_replay_history` as parameters of the learner, since these parameters affect learning. I think the current concept `Actor` is not yet entirely clear to me...

jbrea:
Hi @findmyway 

Amazing progress!

I think we could merge this pretty much now into master. Like this I think it is easier to distribute the workload. I would be happy to tackle some of the open issues.

Regarding your comments above:
1. The new buffers look great. I didn't quite get, however, if our old `getindexconsecutive` is also already implemented. I saw methods to extract `nstep` consecutive buffer entries, but when you think e.g. of DQN on Atari games, one has both the option to choose the `nstep` horizon of bootstrapping and the `kstep` horizon of concatenating subsequent frames to form one state, as e.g. done [here](https://github.com/openai/baselines/blob/7c520852d9cf4eaaad326a3d548efc915dc60c10/baselines/common/atari_wrappers.py#L188).
2. I think, what is called `ActionSelector` now, was called `Policy` before. But I think `ActionSelector` is a great name, since it has probably fewer connotations.
3. The Multi-stage callbacks are a great!
4. I could work on that. Model-based MDP solvers like value iteration are basically just learners.
5. Yes, it would be great to include ReinforcementLearningAnIntroduction.jl in this org!

I have to head off. I'll add some thoughts later.

examples/dqn.jl
```diff
@@ -0,0 +1,41 @@
...
+agent = DQN(learner, buffer, selector;γ=0.99)
+
+function f(n)
+    i = 0
+    (x...) -> begin
+       res = i >= n
+       i += 1
+       res
+    end
+end
```
jbrea: Couldn't one use one of the already defined `stop_conditions` here?

findmyway: Oops, I forgot to update it here.

src/components/agents/q_agent.jl
findmyway: Yeah, I also had the same thought before. I'll explain the decision in detail soon.

findmyway:
I didn't cover the pre-processor part in the picture above. I was planed to have it in the environments.

image

And a `WrappedEnv(env, preprocessor)` will act just like a normal environment. But now I think your comments "Regarding the concept agent..." make more sense. If we treat the agent as an organic whole, then we'd better move the pre-processor part into the agent.

Here's the more detailed one:

image

Let me explain these concepts in detail here:

- **Approximator** acts like a black box here. We feed in the feature and get some values we are interested in(like Q-value, V-value, Advantage...). To update these approximators, we can either provide some gradients (for the neural network approximator) or some corrections (for the tabular approximator).
- **Learner** is a wrapper around **Approximator**, it takes in the training data (an episode of experience or a batch of experiences from model/buffer) and decides how to generate the **corrections**/**gradients/losses** for **Approximator**s. So **it doesn't need to care how the training data are cooked**. It's job is that, hey give me some data in this format and I can use it to update my approximator a little.
    - For example, in the very high level, Q-learner doesn't care much about the `update_horizon`, `γ` or `min_replay_history`. Given `states`, `next_states`, `discounted_rewards`, `terminals`, the learner then calculate the error according to `Q(s) - (r + Q(s') * is_terminal)` and use it to update the approximator.
    - Similar to the double-DQN, we only define `Q(s) - (r + Q'(s') * is_terminal)`, and let the agent prepare the data for it.
    - In theory, we could move the *preparing training data* part into the learner. But I found that keeping each concept as thin as possible can avoid a lot of verbose code by leveraging **multiple-dispatch**.
- **Agent** is a wrapper around **Learner**, it may have an extra **buffer** and/or an **environment model** component. I think **Agent** is very similar to the **learner** concept previously. But it's a kind of combination of different pieces. In the end, we may have many different kinds of agents, and they just act like glue for different components.
- **Actor** is not explicitly named in our code. I prefer to treat it as a conceptual object. When interacting with the environments, the **Agent** just acts like an actor.
    - For Q-value learner based agents, the actor is a combination of *learner* and *action_selector*.
    - For deterministic policy based agents, the policy learner will act like an actor.
    - For simulation based learner, the actor may be something else.
    - So I found it hard to give it a concrete definition and only use **Agent** in the code.

For the `tile-coding`, I treat it as a kind of non-linear tabular approximator here. But it could also be moved to the preprocessor part as you mentioned above. 

Thanks for your reading! I hope I make myself clear above 😄. 
Again, I'm always open for discussions.

<hr>

- For the `getindexconsecutive ` function, yes it's very useful. I removed it temporarily and forgot to move it back🤣

examples/dqn.jl
findmyway: Updated!

jbrea:
Awesome, that's a great summary of your design choices, thanks a lot! The concepts learner and agent make more sense to me now.

Whether the preprocessors are part of the agent or come in form of an environment wrapper, doesn't really matter to me. To use again the analogy with biological agents, one could also say that the sensory organs and early sensory processing are just wrappers of the environment. 

What still confuses me, however, is that conceptually I don't yet really see a difference between e.g. image preprocessing and extracting Fourier/Polynomial/Linear/Tiling features, but now it seems that part of this static preprocessing will be handled by approximators and the rest by environment wrappers. How do I know in the future whether some static preprocessing should go into function approximation or environment wrapping?

findmyway:
Oh, I see.

I'd suggest removing all those static preprocessing parts from the `approximators` folder to avoid confusion. Sorry that I didn't notice this part before. Modifications are coming soon.

findmyway:
I just copied all the preprocessors here😅 Will add some test cases later.

jbrea:
Nice, that looks very elegant to me.
Please feel free to merge this PR whenever you like!

findmyway:
Closing this PR, I merged the changes in https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/26

And let's use https://github.com/orgs/JuliaReinforcementLearning/projects/1 to track the progress.

src/components/agents/q_agent.jl
findmyway: I have been thinking of this comment for a lot of times this week. Now I would like to retract my comment. Fortunately, it's not too late... 😃 

It's true that it would be more intuitive to move those parameters in the learner. This change will be reflected in the latest PR.

Thanks again for your great thoughts!

jbrea: Nice!

PR #39, opened 2019-10-11
add docs

findmyway:


findmyway:
Hi @jbrea ,

Could you help to check why the documentation build step failed on the master branch when you have time?

https://travis-ci.org/JuliaReinforcementLearning/ReinforcementLearning.jl/jobs/599653659#L368-L375

It seems that the ssh key is expired?

jbrea:
Yeah, I think that was it. Should be working now.

findmyway:
Great!

PR #45, opened 2019-10-29
Refactor buffer

findmyway:
## 📝Description

- Update dependency due to upstream change (Project.toml and Manifest.toml)
- Refactor `CircularArrayBuffer` AGAIN. Add a more conventional constructor similar to Array. Access speed is increased a lot!
- Add a benchmark file

## Types of changes

- [x] 🐞Bug fix (non-breaking change which fixes an issue)
- [x] 💪New feature (non-breaking change which adds functionality)
- [x] 🔥Breaking change (fix or feature that would cause existing functionality to change)

## 🚨Pull request checklist

- [ ] My code follows the [Code Style: Blue](https://github.com/invenia/BlueStyle).
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

💔Thank you!


findmyway:
According to my test, the fps of DQN(using the same parameters with those defined in Dopamine) is about 164. The bottleneck is the data transfer from CPU to GPU.

The `consecutive_view` now returns a reshaped view, and then it's further processed by `extract_RTSA` results in a view of reshaped view. According to the benchmark test, it takes about 7ms to turn such an object into CuArray. By using some tricks, it can be reduced to ~2ms. But it will make the code less general. Someone else may take the challenge to come up with a nice solution.

For now, the fps of 164 is not so bad to me.

PR #305, opened 2021-05-23
update the cover saver for demos BasicDQN_CartPole and BasicDQN_EmptyRoom

johnnychen94:
- It seems that Documenter is complaining that `run(ex)` errors, so I work it around by wrapping it as a non-executed version `julia`.
- It seems that `Experiment(..., "")` with empty description will still print a new line, which I think can be improved.

preview:

demo


findmyway:
Closing this in favor of https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/308

Issue #246, opened 2021-04-02
Getting figure object; How to get an animation using GR.plot in CartPolEnv


JinraeKim:
# Notes
- Julia v1.6.0
- ReinforcementLearning.jl v0.8.0
# Summary
I read [lines](https://github.com/JuliaReinforcementLearning/ReinforcementLearningEnvironments.jl/blob/db5a7929f0683709c1a6f7936ee770bad9e78f17/src/environments/examples/CartPoleEnv.jl#L137) and tried to get an animation as described [here](http://docs.juliaplots.org/latest/generated/gr/#gr-ref2) using `Plots.jl`.

However, it seems not to provide a figure object. For example,
```julia
using ReinforcementLearning
using GR
using Plots


function hook_plot_func(anim)
    func_plot = function (t, agent, env)
        GR.plot(env)
        frame(anim)
    end
    hook_plot = DoEveryNStep(func_plot)
end
function test_simple()
    experiment = E`JuliaRL_BasicDQN_CartPole`
    # policy = experiment.policy
    # env = experiment.env
    # s = state(env)
    # @show s
    # a = policy(env)
    # @show a
    anim = Animation()
    hook_plot = hook_plot_func(anim)
    hook = ComposedHook(experiment.hook, hook_plot)
    # hook = ComposedHook(experiment.hook, TotalRewardPerEpisode())
    # hook = TotalRewardPerEpisode()
    run(experiment.policy, experiment.env, experiment.stop_condition, hook)
    gif(anim, "here.gif", fps=15)
    # run(experiment.policy, experiment.env, experiment.stop_condition, experiment.hook)
end

```
and run `test_simple()` says
```juliaREPL
julia> test_simple()
ERROR: No current plot/subplot
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] current
    @ ~/.julia/packages/Plots/z5Msu/src/plot.jl:15 [inlined]
  [3] frame(anim::Animation)
    @ Plots ~/.julia/packages/Plots/z5Msu/src/animation.jl:18
  [4] JuliaReinforcementLearning/ReinforcementLearning.jl#130
    @ ~/.julia/dev/MyPackage/test/rl.jl:9 [inlined]
[...]

PR #309, opened 2021-05-25
add plotting for cartpole and mountaincar with Plots.jl

jamblejoe:
see https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/246

### DIFF
--- a/src/ReinforcementLearningEnvironments/src/ReinforcementLearningEnvironments.jl
+++ b/src/ReinforcementLearningEnvironments/src/ReinforcementLearningEnvironments.jl
@@ -32,6 +32,9 @@ function __init__()
     @require OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed" include(
         "environments/3rd_party/AcrobotEnv.jl",
     )
+    @require Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80" include(
+         "plots.jl",
+    )
 
 
 end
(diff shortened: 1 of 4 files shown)


jamblejoe:
The cartpole and the mountain car look as follows. I am happy to change the appearance when necessary.

grafik

grafik

zsunberg:
I wonder if Plots.jl is the right package to do this with since it is not really a data plot. Maybe something like Compose.jl or Cairo.jl that is focused on general 2D graphics would be better (A compose mountaincar example is here: https://github.com/JuliaPOMDP/QuickPOMDPs.jl/blob/d5c51e68e1a089b1e0fe9518493b27e31cd4c1aa/examples/mountaincar_with_visualization.jl#L18-L27)

jamblejoe:
@zsunberg I agree. The previous implementation used `GR` but did not return a figure object to quickly do animations and it did not work out of the box with jupyter notebooks.

The PR was a fix I initially wrote for myself but was asked to create a PR. I am only familiar with `Plots.jl`, which is using `GR` as backend by default. It gets the job - animations and jupyter support - done quick and dirty. 

I dont know `Compose.jl` or `Cairo.jl`.

findmyway:
I prefer to merge this first.
Rendering is not a core component in RLEnvs.jl and this PR moves one step further to drop the dependency of GR.
Different environments may have many different approaches to render environments. For example, GridWorlds.jl uses Makie, the Atari simply returns a Matrix. At least this PR gives users an option to *plot* the default `CartPole` and `MountainCar` environment.

findmyway:
@all-contributors please add @jamblejoe  for code

PR #321, opened 2021-06-11
update  tips

### LINKED ISSUE #320: Local development environment setup tips causing error
https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/dc54ba44ea955d0d50c23be7c5d06912a0900a86/docs/src/tips.md#L10-L16

Running `dev .` in the folder of the currently active project seems a bit odd, seems like it would create some recursive dependency, and it generates an error for me. 

Was the intention to add the cloned folder to some other environment (i.e. same instructions without the `--project`)? Why not just do `dev ReinforcementLearning` and then you get the whole repo in .julia/dev/ReinforcementLearning and can edit it as you want there?

findmyway:
fix #320

### DIFF
--- a/docs/src/tips.md
+++ b/docs/src/tips.md
@@ -11,15 +11,17 @@ turned into the development mode automatically. The recommended process is:
 1. `cd ReinforcementLearning.jl`
 1. `julia --project`
 1. Press `]` to enter package mode in the Julia REPL.
-1. `pkg> dev .`
 1. `pkg> instantiate`
 1. `pkg> up` (optional)
 
-Then you can modify the code in your favourate editor and test it as usuall.
-Sometimes, you may need to add some other dependencies. Remember to switch the
+Then you can modify the code in your favorite editor and test it as usual.
+Sometimes, you may need to add some extra dependencies. Remember to switch the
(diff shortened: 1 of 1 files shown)


albheim (approved):
Looks good to me. 

Also just tested to create a new empty env and run `dev ReinforcementLearning` and checked the Manifest.toml where it seems tlike the subpackages are linked to the subfolders of .julia/dev/ReinforcementLearning folder. So I think you might not even need to dev subpackages, but can just start changing in those folders (unless you only want a subpackage without the top level one). But I'm not completely clear on how developed vs added packages are treated differently so maybe there is something more to it.

findmyway:
> Looks good to me.
> 
> Also just tested to create a new empty env and run `dev ReinforcementLearning` and checked the Manifest.toml where it seems tlike the subpackages are linked to the subfolders of .julia/dev/ReinforcementLearning folder. So I think you might not even need to dev subpackages, but can just start changing in those folders (unless you only want a subpackage without the top level one). But I'm not completely clear on how developed vs added packages are treated differently so maybe there is something more to it.

Thanks again!
I test it again locally. Surprisingly, you are right here. I'll update the doc correspondingly.

PR #348, opened 2021-07-07
add Base.:(==) and Base.hash for AbstractEnv and test nash_conv on KuhnPokerEnv

peterchen96:
Add Base.:(==) and Base.hash for AbstractEnv so that can compute nash_conv on KuhnPokerEnv().

### DIFF
--- a/src/ReinforcementLearningBase/src/interface.jl
+++ b/src/ReinforcementLearningBase/src/interface.jl
@@ -405,6 +405,10 @@ Make an independent copy of `env`,
 @api copyto!(dest::AbstractEnv, src::AbstractEnv)
 
 # checking the state of all players in env is enough?
+"""
+    Base.:(==)(env1::T, env2::T) where T<:AbstractEnv
+!!! warning: Only check the state of all players in the env.
+"""
 Base.:(==)(env1::T, env2::T) where T<:AbstractEnv = 
     sum([state(env1, player) == state(env2, player) for player in players(env1)]) == length(players(env1))
 Base.hash(env::AbstractEnv, h::UInt) = hash([state(env, player) for player in players(env)], h)
--- a/src/ReinforcementLearningBase/src/interface.jl
+++ b/src/ReinforcementLearningBase/src/interface.jl
@@ -407,10 +407,14 @@ Make an independent copy of `env`,
 # checking the state of all players in env is enough?
 """
     Base.:(==)(env1::T, env2::T) where T<:AbstractEnv
-!!! warning: Only check the state of all players in the env.
+!!! warning
+    Only check the state of all players in the env.
 """
-Base.:(==)(env1::T, env2::T) where T<:AbstractEnv = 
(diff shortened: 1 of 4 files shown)


src/ReinforcementLearningEnvironments/src/environments/examples/examples.jl
```diff
@@ -11,3 +11,7 @@ include("CartPoleEnv.jl")
 include("MountainCarEnv.jl")
 include("PendulumEnv.jl")
 include("BitFlippingEnv.jl")
+
+# checking the state of env is enough?
+Base.:(==)(env1::AbstractEnv, env2::AbstractEnv) = state(env1) == state(env2)
+Base.hash(env::AbstractEnv, h::UInt) = hash(state(env), h)
```
findmyway, line 17: Maybe we can move this into RLBase?

By the way, you can restrict the type to be of the same env type like this:

```suggestion

# checking the state of env is enough?
Base.:(==)(env1::T, env2::T) where T<:AbstractEnv = state(env1) == state(env2)
Base.hash(env::AbstractEnv, h::UInt) = hash(state(env), h)
```

Also add a doc string to warn users that the we only check the state instead of the game specification part.

findmyway (changes requested):

src/ReinforcementLearningEnvironments/src/environments/examples/examples.jl
peterchen96: Thank you for your kind advice. I'll push changes later.

src/ReinforcementLearningBase/src/interface.jl
```diff
@@ -404,6 +404,15 @@ Make an independent copy of `env`,
 @api copy(env::AbstractEnv) = deepcopy(env)
 @api copyto!(dest::AbstractEnv, src::AbstractEnv)
 
+# checking the state of all players in env is enough?
+"""
+    Base.:(==)(env1::T, env2::T) where T<:AbstractEnv
+!!! warning: Only check the state of all players in the env.
```
findmyway, line 410:
```suggestion
!!! warning
    Only check the state of all players in the env.
```

This is the correct format for adding `warning`. Please read https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions for more details.

src/ReinforcementLearningBase/src/interface.jl
```diff
@@ -404,6 +404,15 @@ Make an independent copy of `env`,
...
+"""
+Base.:(==)(env1::T, env2::T) where T<:AbstractEnv = 
+    sum([state(env1, player) == state(env2, player) for player in players(env1)]) == length(players(env1))
```
findmyway, line 413: Use `all` instead of `sum` here. And the list comprehension is not necessary, an iterator will be enough.

peterchen96: Thank you for your responding and I have pushed changes to the branch.

findmyway (approved):

PR #361, opened 2021-07-12
Framework

pilgrimygy:
This is a simulated PR. This content includes:
1. Update the framework.
2. Offline DQN algorithm.
3. Simple CQL loss component. After testing, it can greatly improve the performance of offline DQN.
4. Fix a small bug.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -309,6 +309,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "00krishna",
+      "name": "Krishna Bhogaonker",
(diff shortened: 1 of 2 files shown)


pilgrimygy:
Openspial experiments are not defined because 'tangle=false'.

findmyway:
> Openspial experiments are not defined because 'tangle=false'.

Just remove that test in `RLExperiments` to pass the tests.

By the way, why are the svg files removed?

pilgrimygy:
Mmmmm, I thought it was generated when I ran the experiment myself. I will put it back later.

PR #365, opened 2021-07-12
Add the tuning entropy component

pilgrimygy:
Update the complete implementation and modify some details based on the [SAC](https://arxiv.org/pdf/1812.05905.pdf) paper.


src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
```diff
@@ -40,12 +42,15 @@ end
 - `target_qnetwork2`,
 - `start_policy`,
 - `γ = 0.99f0`,
-- `ρ = 0.995f0`,
+- `τ = 0.005f0`,
```
findmyway, line 45: Any special reason why we switch `ρ` to `τ` here?

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
```diff
@@ -179,11 +193,16 @@ function RLBase.update!(p::SACPolicy, batch::NamedTuple{SARTS})
     end
     update!(p.policy, p_grad)
 
+    # Tune entropy automatically
+    if p.automatic_entropy_tuning
+        p.α -= p.lr_alpha * mean(-log_π .- p.target_entropy)
```
findmyway, line 198: What's the meaning of  `target_entropy` here? It seems to only be set to `Float32(-action_dims)` at the init step.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
Mobius1D: I think what he went for is a standard terminology that is used for soft target updates. I have seen it in a lot of implementations. That is what is used in https://arxiv.org/pdf/2008.10861.pdf as well.

pilgrimygy: @Mobius1D is right. I only change the presentation to match the pseudo-code of the paper.

findmyway: I see. This will be a breaking change. So could you also update https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/master/NEWS.md#reinforcementlearningzoojl and bump the version of https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/c7b66e47e939592045326485b9c89b1486df165d/src/ReinforcementLearningZoo/Project.toml#L4
@pilgrimygy

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
pilgrimygy: In original SAC paper:
[](https://imgtu.com/i/WkqfSJ)
In fact, this loss is linear for α. So I did not use the optimizer.
\bar{\mathcal{H}} is the target entropy, a constant. I initialized it according to the method of the paper:
[](https://imgtu.com/i/WkLGc9)

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
pilgrimygy: OK, I will update it.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
findmyway: Now I see. Thanks for your explanation!

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
findmyway: You need to create a new section named `ReinforcementLearning.jl@v0.10.0`

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
Mobius1D: image
In https://arxiv.org/pdf/1812.05905.pdf it implements automatic temperature tuning and the gradient wrt alpha might be the following expression?
I have seen an implementation like that [here](https://github.com/denisyarats/pytorch_sac/blob/81c5b536d3a1c5616b2531e446450df412a064fb/agent/sac.py#L123)

Mobius1D: Got it

pilgrimygy: Thank you very much. Let me think about it.

pilgrimygy: I think that log \pi and α are irrelevant.

PR #370, opened 2021-07-14
Make general components

pilgrimygy:
Put GaussianNetwork and DuelingNetwork into RLCore as general components. And make a general `evaluate` function.


src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -59,4 +60,63 @@ RLBase.update!(app::ActorCritic, gs) = Flux.Optimise.update!(app.optimizer, para
...
+
+function (m::GaussianNetwork)(S)
+    x = m.pre(S)
+    m.μ(x), m.logσ(x) 
+end
+
+"""
+This function is compatible with a multidimensional action space.
+"""
+function evaluate(model::GaussianNetwork, state)
```
findmyway, line 92: The function name `evaluate` seems to be too general to me. Maybe we can simply use functional call here?

- Also, I think we can add a keyword argument `is_return_log_prob=true` to control whether to return `logp_π` or not. I think in some cases, we only need the actions.
- Doc string should also be updated to make it clear that the `tanh` is applied.
- A positional argument `rng::AbstractRNG` should also be added to ensure reproducibility.

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -1,6 +1,7 @@
-export NeuralNetworkApproximator, ActorCritic
+export NeuralNetworkApproximator, ActorCritic, GaussianNetwork, evaluate, DuelingNetwork
```
findmyway, line 1: Remove `evaluate` when we use callable object in `GausianNetwork`.

findmyway (changes requested):
1. Remember to update `NEWS.md`.
2. The failure in CI seems to be unrelated to this PR. I'll take a look into it.

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
pilgrimygy: Well. I will complete it.

pilgrimygy: What does `functional call` means? @findmyway

pilgrimygy: In addition, do I need to update the version of RLCore?

findmyway:
```
(model::GaussianNetwork)(state; is_return_log_prob=true) = ...
(model::GaussianNetwork)(rng, state; kw...) = model(GLOBAL_RNG, state;kw...)
```

pilgrimygy: Doesn't this conflict with the following code?
```julia
function (m::GaussianNetwork)(S)
    x = m.pre(S)
    m.μ(x), m.logσ(x) 
end
```

findmyway: Hmm, you are right. Can these two functions be merged somehow?Or just give it a more specific name?

pilgrimygy: Maybe we can use this form:
```julia
(model::GaussianNetwork)(state, is_return_log_prob=true; rng) = ...
```

src/ReinforcementLearningZoo/Manifest.toml
```diff
@@ -456,7 +456,7 @@ version = "0.9.6"
 deps = ["AbstractTrees", "Adapt", "CUDA", "CircularArrayBuffers", "Compat", "Dates", "Distributions", "ElasticArrays", "FillArrays", "Flux", "Functors", "GPUArrays", "LinearAlgebra", "MacroTools", "Markdown", "ProgressMeter", "Random", "ReinforcementLearningBase", "Setfield", "Statistics", "StatsBase", "UnicodePlots", "Zygote"]
 path = "../ReinforcementLearningCore"
 uuid = "de1b191a-4ae0-4afa-a27b-92d07f46b2d6"
-version = "0.8.1"
+version = "0.8.2"
```
findmyway, line 459: Did you manually edit this file?
You only need to update this line  https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/master/src/ReinforcementLearningZoo/Project.toml#L37

and set it to `0.8.2` instead of `0.8`. Then run `]resolve` and this file will be automatically updated.

findmyway (changes requested):

src/ReinforcementLearningZoo/Manifest.toml
pilgrimygy: No, I only change the RLcore version and run `]resolve`.

findmyway: I see. Please update the compat of `RLCore` as I said above, so that it will not accidentally rely on an old version of `RLCore`.

pilgrimygy: Well. I will update it.

findmyway:
It's quite strange that the tests now run much slower...

pilgrimygy:
Thanks @findmyway for finding so many bugs!

findmyway:
@pilgrimygy could you confirm that the `JuliaRL_BasicDQN_CartPole` experiment in this PR can be finished in less than a minute?

It seems everything works as usual on my local machine. Not sure why it is so slow in CI.

pilgrimygy:
Well. Why `JuliaRL_BasicDQN_CartPole` in CI is too slow?

findmyway:
I must say, this is the weirdest thing I've seen in Julia so far. And I really don't know why...

```julia
(@v1.6) pkg> activate src/ReinforcementLearningExperiments/

(ReinforcementLearningExperiments) pkg> build

(ReinforcementLearningExperiments) pkg> test

┌ Warning: `InplaceableThunk(t::Thunk, add!)` is deprecated, use `InplaceableThunk(add!, t)` instead.
│   caller = ip:0x0
└ @ Core :-1

```

Then it is stuck. Just like what we see in the CI.

Well, if we run it in REPL:

```julia

(@v1.6) pkg> activate src/ReinforcementLearningExperiments/

julia> using ReinforcementLearningExperiments

julia> ex = E`JuliaRL_BasicDQN_CartPole`;

julia> run(ex)
```

Things work as usual.

PR #374, opened 2021-07-16
Change the installation for gym

Mobius1D:
Change installation for gym from pip to conda

### DIFF
--- a/src/ReinforcementLearningEnvironments/src/environments/3rd_party/gym.jl
+++ b/src/ReinforcementLearningEnvironments/src/environments/3rd_party/gym.jl
@@ -1,5 +1,5 @@
 using .PyCall
-
+using Conda
 # TODO: support `seed`
 function GymEnv(name::String)
     if !PyCall.pyexists("gym")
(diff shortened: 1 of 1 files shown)


src/ReinforcementLearningEnvironments/src/environments/3rd_party/gym.jl
```diff
@@ -1,5 +1,5 @@
 using .PyCall
-
+using Conda
```
findmyway, line 2: Note you added another extra dependency of `Conda` here. You'd better do something similar to https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/dd17ad674f6d8b9935d2a7118d5f4ef1a83751c3/src/ReinforcementLearningEnvironments/src/ReinforcementLearningEnvironments.jl#L22-L24

findmyway (changes requested):

src/ReinforcementLearningEnvironments/src/environments/3rd_party/gym.jl
findmyway: To me this is not a very important function, I'm hesitant to merge it right now because some people may still prefer to use `pip` instead of `conda`. Installing `Conda.jl` on Windows is quite painful... 😢

pilgrimygy: well. I use pip instead of conda. And I use Windows..hhhh

Mobius1D: Huh, I understand. I never have used conda with windows.

But still, pycall uses a version of miniconda to handle python packages internally, so I dont think there would be a problem for users. Plus pycall depends on conda for its package management and it is recommended that Conda.jl is used. Refer 3rd paragraph [here](https://github.com/JuliaPy/PyCall.jl#installation).

If you still think its a problem. Then we can drop this PR.

findmyway: > pycall uses a version of miniconda to handle python packages internally

That's not true. You can configure `PyCall` to use your preferred Python environment.

Mobius1D: Yes, You can do that. I meant by default. 

Alright If it would be a problem with windows users then I would drop the PR.

findmyway: I think we can add this feature later. To me it doesn't worth adding another dependency of `Conda` right now.

Mobius1D: Got it

PR #375, opened 2021-07-16
WIP: add `nfsp` algorithm and relative experiment

peterchen96:
add `nfsp` algorithm and test it on julia version `KuhnPokerEnv`, however, the result is not well for now. #342


docs/experiments/experiments/NFSP/JuliaRL_NFSP_KuhnPoker.jl
```diff
@@ -0,0 +1,88 @@
+"""
+NFSP agents trained on Kuhn Poker game.
+"""
```
johnnychen94, line 3: You can follow this format here: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/fc2afa7b9f78c97fc688dbe3aaed9d4d5de366f3/docs/experiments/experiments/CFR/JuliaRL_DeepCFR_OpenSpiel.jl#L1-L7

This format is defined in in [DemoCards](https://johnnychen94.github.io/DemoCards.jl/stable/quickstart/usage_example/julia_demos/1.julia_demo/#juliademocard_example)

docs/experiments/experiments/NFSP/JuliaRL_NFSP_KuhnPoker.jl
```diff
@@ -0,0 +1,88 @@
...
+pyplot() #hide
+ex = E`JuliaRL_NFSP_KuhnPoker`
+run(ex)
+plot(recorder.episode, recorder.results, xaxis=:log, yaxis=:log)
+xlabel!("episode")
+ylabel!("nash_conv")
+
+savefig("assets/JuliaRL_NFSP_KuhnPoker.png")#hide
+
+# ![](assets/JuliaRL_NFSP_KuhnPoker.png)
```
johnnychen94, line 88: These are used to generate the progress plot. Whether it is needed depends on how you write the `cover` field in the front matter (see my previous comment).

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
+"""
+Neural Fictitious Self-Play (NFSP) agent implemented in Julia.
+
+See the paper https://arxiv.org/abs/1603.01121 for more details.
+"""
+export NFSPAgent, NFSPAgents
```
johnnychen94, line 7: I don't think users need to directly use `NFSPAgent`, so no need to export this one.

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
...
+
+using Distributions: TruncatedNormal
+
+mutable struct NFSPAgents <: AbstractPolicy
+    agents::Dict{Any, AbstractPolicy}
+end
```
johnnychen94, line 13: We don't use plural case `NFSPAgents` to differentiate with `NFSPAgent`.

Since we don't really need to export `NFSPAgent`, we can maybe rename `NFSPAgent` to `NFSPAgentInstance`?

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
+"""
+Neural Fictitious Self-Play (NFSP) agent implemented in Julia.
+
+See the paper https://arxiv.org/abs/1603.01121 for more details.
+"""
```
johnnychen94, line 5: I believe this docstring should go to `NFSPAgents`. Please follow the Julia guideline https://docs.julialang.org/en/v1/manual/documentation/

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
...
+mutable struct NFSPAgents <: AbstractPolicy
+    agents::Dict{Any, AbstractPolicy}
+end
+
+mutable struct NFSPAgent <: AbstractPolicy
+    η
+    rng
+    rl_agent::Agent
+    sl_agent::Agent
+end
```
johnnychen94, line 20: Is this a usable policy? Otherwise there's no need to `<: AbstractPolicy`.

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
...
+Neural Fictitious Self-Play (NFSP) agent implemented in Julia.
+
+See the paper https://arxiv.org/abs/1603.01121 for more details.
+"""
+export NFSPAgent, NFSPAgents
+
+using Distributions: TruncatedNormal
+
+mutable struct NFSPAgents <: AbstractPolicy
+    agents::Dict{Any, AbstractPolicy}
```
johnnychen94, line 12: Just a question: does it really accept any type of `AbstractPolicy`?

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
```diff
@@ -0,0 +1,190 @@
...
+end
+
+mutable struct NFSPAgent <: AbstractPolicy
+    η
+    rng
+    rl_agent::Agent
+    sl_agent::Agent
+end
+
+function initW(out_size, in_size)
```
johnnychen94, line 22: `initW` is so common a name that might cause conflict with other algorithms.

```suggestion
function _NFSP_initW(out_size, in_size)
```

johnnychen94 (commented):
random comments from random person :)

src/ReinforcementLearningZoo/src/algorithms/nfsp/nfsp.jl
peterchen96: Thanks for your reviews and sorry for the late response.

>> I don't think users need to directly use `NFSPAgent`, so no need to export this one.
>
>> We don't use plural case `NFSPAgents` to differentiate with `NFSPAgent`.
> >
> >Since we don't really need to export `NFSPAgent`, we can maybe rename `NFSPAgent` to `NFSPAgentInstance`?

Here, `NFSPAgent` is the core policy to play the game. In my view, agents in one multi-agent experiment may choose different policies to play the game. (maybe just like players can use different schemes in one game) 

And `NFSPAgents` is more like a special [MultiAgentManager](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/639717388fb41199c98b90406bea76232bc6294d/src/ReinforcementLearningCore/src/policies/agents/multi_agent.jl#L8) in which all agents choose `nfsp` policy to learn the best response. For more clarity, I'll extract 'NFSPAgents' and its relative methods to a new file and renamed it as 'NFSPAgentManager'.

>  >You can follow this format here:
>
> > I believe this docstring should go to `NFSPAgents`. Please follow the Julia guideline https://docs.julialang.org/en/v1/manual/documentation/

Thanks for your reminder. I'll supplement the description about `NFSPAgent` and its experiment later.

> `initW` is so common a name that might cause conflict with other algorithms.

Thank you for pointing it out. Maybe I'll rename it as `_TruncatedNormal` because of the used distribution.

findmyway (changes requested):
The implementation seems to be totally incorrect. I'll sync with you offline later.

peterchen96:
Sorry for late response. Since this PR’s implementation was totally incorrect, I would close this and create a new PR to record my implementation and relative experiment.

PR #407, opened 2021-07-27
WIP: CRR algorithm

pilgrimygy:
@findmyway 
I finish a simple `OfflinePolicy` framework and we can use it to call the online algorithms by offline dataset.
I test the discrete CRR that has good performance.
But the continuous CRR is poor. I don't know why. I guess the `actor_loss_coef` calculation is wrong.


pilgrimygy:
And I test the Distributional Critic is inferior to normal Critic in performance.

PR #414, opened 2021-07-28
Reduce allocations and update docstring for GaussianNetwork

albheim:
PR Checklist

- [ ] Update NEWS.md?

Found a spot where the whole expression wasn't dotted so there were still some unnecessary allocations in it. Also saw the docstring was a bit inconsistent with showing default values so added that in also.

### DIFF
--- a/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
+++ b/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
@@ -87,7 +87,7 @@ This function is compatible with a multidimensional action space. When outputtin
 
 - `rng::AbstractRNG=Random.GLOBAL_RNG`
 - `is_sampling::Bool=false`, whether to sample from the obtained normal distribution. 
-- `is_return_log_prob`, whether to calculate the conditional probability of getting actions in the given state.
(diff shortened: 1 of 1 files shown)


findmyway:
Any idea why the CI failed?

findmyway:
> Btw, is there a reason to not have a test in the main repo that runs all the tests like the ci does? Would simplify just running all the tests locally, but I guess it is not that hard to just copy the code from the ci script as long as you know where to find it.

The ci simply runs tests in separate subpackages:

https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/472bf301171da54de8802b47d8a7c811b2f6c6a6/.github/workflows/ci.yml#L48-L64

albheim:
> The ci simply runs tests in separate subpackages:

Yes, exactly. I copied that code to run it locally. But what I'm saying is that if one wants to run tests locally it would be nice and easy to just have to write `test ReinforcementLearning` and it would run all of that for you while still keeping the option to run the packages indivudually. But I'm not sure if this is even possible to do in a good way?

pilgrimygy:
Thank you very much~

findmyway:
I see.

>  But I'm not sure if this is even possible to do in a good way?

I'm not sure how to do it 😆 

I used to use https://github.com/nektos/act to run github actions locally. But it now has a problem in this repo.

PR #415, opened 2021-07-28
Fix a bug

pilgrimygy:
@rbange There seems to be a problem with the code, I helped modify it.


findmyway:
Thanks!

@all-contributors please add @pilgrimygy for bug

PR #426, opened 2021-07-30
Add wrapped_env[!] to access env inside wrappers

albheim:
Have not added any test or documentation for this, should I?


albheim:
This is a convenience method to access the env beneath layers of wrappers. It will iterate through all `AbstractEnvWrapper`s until it reaches an `AbstractEnv`. It will not work with for example `MultiThreadEnv` since that is outside the wrappers.

findmyway:
Cool!

PR #429, opened 2021-08-03
Add Atari datasets released by Google Research

Mobius1D:
This is a draft PR for loading Atari datasets. 

PR Checklist

- [x] Update NEWS.md?
- [x] Test atari_datasets
- [x] Update docs
- [x] Fix spell check


src/ReinforcementLearningDatasets/src/atari/npzreadarray.jl
```diff
@@ -0,0 +1,188 @@
+# Credits: https://github.com/fhs/NPZ.jl
```
findmyway, line 1: Why do we need this file?

Mobius1D: The dataset is stored in numpy array files. I did not know how to extract it. So, NPZ.jl provides a neat package to manage this extraction, but it supports input as the string of the file path(.npy or .npz) or IO. It does work on GZStream but is not supported. 

So this is just a temporary workaround. I am planning to make a PR in that repo to make it work by just importing NPZ.jl

Mobius1D: It defines readnpzarray function that is used. I did not put all the functions in, so it is not working. Will add everything and it will show in the tests.

src/ReinforcementLearningDatasets/src/atari/atari_dataset.jl
```diff
@@ -0,0 +1,169 @@
+"""
+Represents an iterable dataset of type AtariDataSet with the following fields:
+
+`dataset`: Dict{Symbol, Any}, representation of the dataset as a Dictionary with style as `style`
+`epochs`: Vector{Int}, list of epochs to load
+`repo`: String, the repository from which the dataset is taken
+`size`: Integer, the size of the dataset
```
findmyway, line 7: the meaning of size is unclear to me. change another name?

src/ReinforcementLearningDatasets/src/atari/atari_dataset.jl
```diff
@@ -0,0 +1,169 @@
...
+    else
+        if (state+1) * batch_size <= size
+            inds = state*batch_size+1:(state+1)*batch_size
+        else
+            return nothing
+        end
+        state += 1
+    end
+
+    batch = (state = copy(ds.dataset[:state][:, :, inds]),
```
findmyway, line 147: I think we don't need copy here?

src/ReinforcementLearningDatasets/src/atari/npzreadarray.jl
findmyway: so based on your description, what we need is to open a gzip file as IO stream and then use NPZ.jl to extract data, right?
If so I remember there's a package to do the uncompression

findmyway (approved):
LGTM

src/ReinforcementLearningDatasets/src/atari/atari_dataset.jl
Mobius1D: Yep. I just wanted to give the user the flexibility to change the value of the batches.

I haven't come across such a use. Will change it.

src/ReinforcementLearningDatasets/src/atari/npzreadarray.jl
Mobius1D: NPZ will take care of just the reading part and not the decompression.

So the plan here is to decompress using gzipopen and use NPZ to read from the stream.

Mobius1D: I am looking at TranscodingStreams.jl for now.

Mobius1D: Works

src/ReinforcementLearningDatasets/src/atari/atari_dataset.jl
Mobius1D: size denotes the size of the dataset, It seemed clear to me. Would be great if you could suggest something better.

Something like dataset_size?

findmyway: How about `length` based on the following line:

https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/429/files/ca62405e2247e7c5ab9ac0f0b1caa0f16d94255f#diff-ee59e3b8adc8856b61bbc10809639b624390891bbd4366bc262405dc63a1964fR161

Mobius1D: yep, makes sense

findmyway:
I think we can merge it once the CI error is resolved.

PR #439, opened 2021-08-06
correct `nfsp` implementation

peterchen96:
The current implementation of `nfsp` is incorrect since it errored processing some train samples' information for `rl_agent` in some stages.

Here, I correct the update process for `NFSPAgent,` and the experiment's result looks stable for now when the number of train episodes is larger than the epsilon decay step.


findmyway:
@peterchen96  Can we decrease the runtime of NFSP?

Before adding the NFSP experiment, the document CI can be finished in 40m. Now it takes over an hour.

PR #440, opened 2021-08-07
update the nfsp experiment's parameters

peterchen96:
#439 mentioned that the NFSP experiment's running time is too long.

I've tried to update the experiment's parameters to decrease the running time.

For now, the running time is about 5 min, and the performance looks well.


findmyway:
Thanks!

PR #447, opened 2021-08-11
try to fix bugs of ActionTransformedEnv

peterchen96:
When I want to get the action space of the player 1 with an `ActionTransformedEnv,` it will raise errors like the following:

```Julia
julia> using ReinforcementLearning
julia> env = KuhnPokerEnv();
julia> wrapped_env = ActionTransformedEnv(env);
julia> action_space(wrapped_env, 1)
ERROR: MethodError: no method matching identity(::Base.OneTo{Int64}, ::Int64)
```

Maybe we should move `args...` into the `action_space` function...


findmyway:
@all-contributors please add @peterchen96  for code, doc

findmyway:
Sorry, it seems I forgot to add you in the README.

PR #454, opened 2021-08-14
disable notebook generation

### LINKED ISSUE #421: Links of  `show nbview` badges in RLExperiments are incorrect
and most `show nbview` links in the experiments are broken (sorry I am reporting these as I find them hence the separate comments)

_Originally posted by @Krastanov in https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/418#issuecomment-888510224_

Watch: https://github.com/johnnychen94/DemoCards.jl/issues/97#issue-955482099

johnnychen94:
This would disable all notebook generation and thus reduce the generation time from 3x to 2x. A side effect is #421 is no longer an issue (since there's no nbview anymore 😆 )

This requires the master branch of DemoCards (https://github.com/johnnychen94/DemoCards.jl/pull/98) to actually work. Currently, it should not change anything.

It's up to you to decide whether you want to merge it now or then. Both are valid choices from my understanding.

closes #421

### DIFF
--- a/docs/experiments/config.json
+++ b/docs/experiments/config.json
@@ -1,3 +1,6 @@
 {
-    "theme": "grid"
+    "theme": "grid",
+    "properties":{
+        "notebook": "false"
+    }
 }


findmyway:
Thanks!

PR #480, opened 2021-08-23
FisherBRC algorithm and update docs

pilgrimygy:
In this PR:
1. Implement FisherBRC algorithm. I test it in Pendulum env, and FisherBRC's performance is better than online SAC. (But it is very slow)
2. Update docs, including CRR, PLAS, and SAC.
3. Fix some bugs.


findmyway (approved):
LGTM

PR #486, opened 2021-08-27
add GPU support for GaussianNetwork, fix #455

### LINKED ISSUE #455: The problem of using GaussianNetwork in gpu
When we test the GaussianNetwork in gpu, after my test, there will be errors in the actions sampled from the standard distribution:
```julia
function (model::GaussianNetwork)(rng::AbstractRNG, state; is_sampling::Bool=false, is_return_log_prob::Bool=false)
    x = model.pre(state)
    μ, raw_logσ = model.μ(x), model.logσ(x) 
    logσ = clamp.(raw_logσ, log(model.min_σ), log(model.max_σ))
    if is_sampling
        π_dist = Normal.(μ, exp.(logσ))
        z = rand.(rng, π_dist)
        if is_return_log_prob
            logp_π = sum(logpdf.(π_dist, z) .- (2.0f0 .* (log(2.0f0) .- z .- softplus.(-2.0f0 .* z))), dims = 1)
            return tanh.(z), logp_π
        else
            return tanh.(z)
        end
    else
        return μ, logσ
    end
end
```
The error message is:
```
LoadError: InvalidIRError: compiling kernel broadcast_kernel(CUDA.CuKernelContext, CUDA.CuDeviceMatrix{Distributions.Normal{Float32}, 1}, Base.Broadcast.Broadcasted{Nothing, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, CUDA.var"#285#286"{Distributions.Normal}, Tuple{Base.Broadcast.Extruded{CUDA.CuDeviceMatrix{Float32, 1}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}, Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{2}, Nothing, typeof(exp), Tuple{Base.Broadcast.Extruded{CUDA.CuDeviceMatrix{Float32, 1}, Tuple{Bool, Bool}, Tuple{Int64, Int64}}}}}}, Int64) resulted in invalid LLVM IR
Reason: unsupported dynamic function invocation (call to string(a::Union{Char, SubString{String}, String}...) in Base at strings/substring.jl:208)
```

burmecia:
`Normal()` distribution doesn't support GPU yet, this patch is a workaround to this issue. It can be reverted once Distributions.jl added GPU support. Watch on [this issue](https://github.com/JuliaStats/Distributions.jl/issues/1067)


findmyway:
Thanks!

PR #490, opened 2021-08-30
correction: two state variables

ShuhuaGao:
PR Checklist

- [ ] Update NEWS.md?

See discussion https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/discussions/475#discussioncomment-1210348


findmyway:
After taking a close look into the implementation of `PendulumEnv` in openai/gym and the one in our package, I realized that the internal state is of length 2 and the observation is of length 3.

https://github.com/openai/gym/blob/master/gym/envs/classic_control/pendulum.py#L51
```
        self.state = np.array([newth, newthdot])
```

https://github.com/openai/gym/blob/master/gym/envs/classic_control/pendulum.py#L58-L62
```
    def _get_obs(self):
        theta, thetadot = self.state
        return np.array([np.cos(theta), np.sin(theta), thetadot], dtype=np.float32)
```

In our package, we simply used `state` everywhere, which is the root cause of your confusion in https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/discussions/475#discussioncomment-1210348 . So I'd take back what I said that this was a bug here. Really sorry!

<hr>

Considering that the API of this package has changed a lot, a more formal definition of the `PendulumEnv` today is to define two sets of states, one for the internal state (of length 2) and another one for the observation state (of length 3). And making the later as the default one.

ShuhuaGao:
Noted with thanks. It makes full sense that the observation can differ from the state. The current implementation is confusing mainly because I am not familiar with OpenAI Gym. Since Gym is currently the de facto standard, it is reasonable to follow their implementations. 

> more formal definition of the PendulumEnv today is to define two sets of states, one for the internal state (of length 2) and another one for the observation state (of length 3).


The above change is indeed meaningful.

PR #492, opened 2021-08-31
add vmpo algorithm

burmecia:
This PR is to add a new policy gradient algorithm [VMPO](https://arxiv.org/abs/1909.12238) from DeepMind, a CartPole experiment based on it is also added.


src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
```diff
@@ -0,0 +1,235 @@
...
+- `γ = 0.99f0`: discount factor
+- `ϵ_η = 0.02f0`: temperature η hyperparameter
+- `ϵ_α = 0.1f0`: Lagrange multiplier α (discrete) hyperparameter
+- `ϵ_αμ = 0.005f0`: Lagrange multiplier α_mu (continuous) hyperparameter
+- `ϵ_ασ = 0.00005f0`: Lagrange multiplier α_σ (continuous) hyperparameter
+- `n_epochs = 8`: update policy for n epochs
+- `dist = Categorical`: `Categorical` - discrete, `Normal` - continuous
+- `rng = Random.GLOBAL_RNG`
+
+By default, `dist` is set to `Categorical`, which means it will only works
```
findmyway, line 31:
```suggestion
By default, `dist` is set to `Categorical`, which means it will only work
```

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
```diff
@@ -0,0 +1,235 @@
...
+            # loss for temperature η
+            Lη = η * p.ϵ_η + η * log(mean(exp.(top_k_advs / η)))
+
+            # loss for Lagrange multiplier α
+            Lα = if is_discrete
+                π_pairs = ignore() do
+                    zip(eachcol(π_old), eachcol(π))
+                end
+                KL = [Flux.kldivergence(i...) for i in π_pairs]
+                mean(α * (p.ϵ_α .- dropgrad(KL)) + dropgrad(α) * KL)
```
findmyway, line 213: The usage of `Flux.kldivergence` is kind of weird to me. I think it supports 2d array, right? Here it seems you only need to set the `agg=identity`.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
```diff
@@ -0,0 +1,235 @@
...
+                KLσ = 0.5f0 * ((σ_old ./ σ) .^ 2 .- 1 .+ 2 * log.(σ ./ σ_old))
+                Lασ = mean(ασ * (p.ϵ_ασ .- dropgrad(KLσ)) + dropgrad(ασ) * KLσ)
+                Lαμ + Lασ
+            end
+
+            loss = Lᵥ + Lπ + Lη + Lα
+        end
+
+        # skip if gradient contains NaN or Inf
+        if any(x -> !isnothing(x) && any(y -> isnan(y) || isinf(y), x), gs)
```
findmyway, line 226: I suppose this should be an unexpected case. Better to throw a warning here?

findmyway (approved):
I haven't look into the details of the original paper yet. But this implementation seems pretty good to me.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
burmecia: Yes, we can add warning here but it maybe too verbose because this can happen many times.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
burmecia: That's valid, I've made the change. Thanks.

burmecia:
It seems http://rail.eecs.berkeley.edu/ is down now, which caused Dataset testing failure.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
findmyway: make sense

findmyway:
Thanks for your contribution!

PR #509, opened 2021-09-22
cpu and gpu test

pilgrimygy:
PR Checklist

- [ ] Update NEWS.md?


findmyway (approved):
LGTM.

By the way, some models are followed by `|> cpu` explicitly. Are you meant to do so?

pilgrimygy:
> LGTM.
> 
> By the way, some models are followed by `|> cpu` explicitly. Are you meant to do so?

Yes, I mean that these algorithms (except BCQD and CRR-discrete) can run on CPU and GPU. In actual operation, we only uses CPU.

PR #510, opened 2021-09-23
Fix dispatch for is_discrete_space

johannes-fischer:
`is_discrete_space(1:5)` was not working for me (when I was trying to convert an environment with `discrete2standard_discrete`).

The reason is that 
`typeof([1,2,3]) isa Type{AbstractVector} == false`
`typeof([1,2,3]) isa Type{<:AbstractVector} == true`

Or from the julia manual: https://docs.julialang.org/en/v1/manual/types/#man-typet-type

> `isa(A, Type{B})` is true if and only if A and B are the same object and that object is a type 

PR Checklist

- [ ] Update NEWS.md?


findmyway:
Thanks!

@all-contributors please add @johannes-fischer  for code

PR #515, opened 2021-09-28
WIP to implement FQE

Mobius1D:
PR Checklist

- [ ] Update NEWS.md?

This is the implementation of FQE on RLZoo. The performance is not as expected and I can't seem to figure out the reason but I suspect that it is because I have not implemented it properly for OfflineRL Framework in RL.jl. 

The exact same algorithm was implemented from a scratch without RL.jl framework and it works as expected. The exact same update function is used in the PR but it doesn't work.

It would be great if I could get your thoughts on it @findmyway @pilgrimygy.
## Details
The code for the implementation that works is given below.

```julia
@showprogress for step in 1:20000
    s, a, r, t, s′ = sample(rng, ds, 256)[2]
    s, a, r, t, s′ = cu.([s, a, r, t, s])
    q′ = Qₜ(vcat(s′, policy(s′)[1])) |> vec
    
    target = r .+ γ .* (1 .- t) .* q′
    gs = gradient(params(Q)) do
        q = Q(vcat(s, reshape(a, :, 256))) |> vec
        loss = loss_func(q, target)
        Zygote.ignore() do
#             @info loss
            push!(losses, loss)
        end
        loss
    end
    Flux.Optimise.update!(optimiser, params(Q), gs)
    if step % tar_update_freq == 0
        Qₜ = deepcopy(Q)
    end
end
```
I do understand that the code has type instability but it's just a prototype.
## Results
```julia
acc = 0.0

for i in 1:100

    reset!(env)

    s = state(env)

    a = agent(env)

    input = vcat(s, a)

    input = cu(input)

    acc += Vector(Qₜ(input))[]
end

acc / 100
```
-232.72509939193725 and it does stay between -230 to -300 similar to the results that we get with  actual tests.

image

The more accurate method would be to have the same initial states of OPE and Online Evaluation but I am pretty sure that the algorithm works based on a number of tests.

image


findmyway:
It's unclear to me why it performs badly.
I'll work on investigating it later after the long vacation in Oct. To unblock your final report, we can merge it first since this work is kind of independent.

Mobius1D:
Sure!

PR #516, opened 2021-09-29
OSPP Report for RLDatasets.jl

Mobius1D:
PR Checklist

- [ ] Update NEWS.md?

This is the report which I would update based on the WIP on FQE PR getting fixed.
I believe that there are still a few spelling mistakes but apart from that the report is mostly finalised. 

Please do have a look and suggest changes if possible @findmyway .


findmyway:
Is it possible to merge it with the mid-term report?

PR #519, opened 2021-10-04
Update reward wrappers to be more consistent

albheim:
PR Checklist

- [ ] Update NEWS.md?

Since there is `ActionTransformedEnv` and `StateTransformedEnv` which does something to the action resp state to transform it, the current `RewardOverriddenEnv` works in much the same way, and should maybe be called `RewardTransformedEnv` and have a more similar constructor. All these get the information action/state/reward and should output a modified version of the input, i.e. transform seems like a good word.

`RewardOverriddenEnv` could instead be used for a case where you want to completely override the existing reward, and need information from the environment to define this (it is not enough to know the normal reward to define the new reward, so it is not merely a transformation of the reward). So instead of being called on the current reward, it is called with the current environment as argument.

This is just a small inconsistency I noticed when wanting the behaviour of my suggestion for `RewardOverriddenEnv`. I think consistency should be valued so at least the name change is worth keeping. The `RewardOverriddenEnv` is otherwise easy enough to implement on your own, but at the same time I think it can be useful often enough for it to stay.

I would be breaking and maybe confusing since if people are using `RewardOverriddenEnv` will not fail on that call now since it is added with different behaviour. But it will at least be most likely to error later on when trying to to scalar operations on the `env` instead of giving strange logical errors.


src/ReinforcementLearningEnvironments/src/environments/wrappers/RewardOverriddenEnv.jl
```diff
@@ -3,12 +3,12 @@ export RewardOverriddenEnv
 """
     RewardOverriddenEnv(env, f)
 
-Apply `f` on `reward(env)`.
+Apply `f` on the current environment to generate a custom reward.
```
albheim, line 6:
```suggestion
Apply `f` on `env` to generate a custom reward.
```

PR #549, opened 2021-11-18
Make experiments GPU compatible

findmyway:
The main idea is that we need to transfer data from CPU to GPU earlier, otherwise `SubArray` and `ReshapedArray` are not well handled on GPU.

### DIFF
--- a/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/A2C.jl
+++ b/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/A2C.jl
@@ -60,15 +60,15 @@ function _update!(learner::A2CLearner, t::CircularArraySARTTrajectory)
     w₃ = learner.entropy_loss_weight
     to_device = x -> send_to_device(device(AC), x)
 
-    states = select_last_dim(t[:state], 1:n)
-    states_flattened = flatten_batch(states) |> to_device # (state_size..., n_thread * update_freq)
+    S = t[:state] |> to_device
+    states = select_last_dim(S, 1:n)
+    states_flattened = flatten_batch(states) # (state_size..., n_thread * update_freq)
 
     actions = select_last_dim(t[:action], 1:n)
     actions = flatten_batch(actions)
     actions = CartesianIndex.(actions, 1:length(actions))
 
-    next_state_values =
-        t[:state] |> select_last_frame |> Array |> to_device |> AC.critic |> send_to_host
+    next_state_values = S |> select_last_frame |> AC.critic |> send_to_host
 
     gains =
         discount_rewards(
--- a/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/ppo.jl
+++ b/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/ppo.jl
@@ -223,17 +223,24 @@ function _update!(p::PPOPolicy, t::AbstractTrajectory)
     w₂ = p.critic_loss_weight
     w₃ = p.entropy_loss_weight
     D = device(AC)
+    to_device(x) = send_to_device(D, x)
 
     n_envs, n_rollout = size(t[:terminal])
     @assert n_envs * n_rollout % n_microbatches == 0 "size mismatch"
     microbatch_size = n_envs * n_rollout ÷ n_microbatches
 
     n = length(t)
-    states_plus = send_to_device(D, t[:state])
+    states_plus = to_device(t[:state])
 
-    states_flatten = flatten_batch(select_last_dim(states_plus, 1:n))
+    if t isa MaskedPPOTrajectory
+        LAM = to_device(t[:legal_actions_mask])
+    end
+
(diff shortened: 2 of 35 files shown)


src/ReinforcementLearningZoo/src/algorithms/policy_gradient/A2C.jl
```diff
@@ -60,15 +60,15 @@ function _update!(learner::A2CLearner, t::CircularArraySARTTrajectory)
     w₃ = learner.entropy_loss_weight
     to_device = x -> send_to_device(device(AC), x)
 
-    states = select_last_dim(t[:state], 1:n)
-    states_flattened = flatten_batch(states) |> to_device # (state_size..., n_thread * update_freq)
+    S = t[:state] |> to_device
+    states = select_last_dim(S, 1:n)
```
albheim, line 64: Not related to this PR specifically, but why do we do `select_last_dim` with `1:n`, shouldn't that be the whole range?

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/ppo.jl
```diff
@@ -242,37 +249,39 @@ function _update!(p::PPOPolicy, t::AbstractTrajectory)
...
             else
                 lam = nothing
             end
-            s = send_to_device(D, select_last_dim(states_flatten, inds))  # !!! performance critical
-            a = send_to_device(D, select_last_dim(actions_flatten, inds))
+
+            # s = to_device(select_last_dim(states_flatten_on_host, inds))
+            # !!! we need to convert it into a continuous CuArray otherwise CUDA.jl will complain scalar indexing
+            s = to_device(collect(select_last_dim(states_flatten_on_host, inds)))
+            a = to_device(collect(select_last_dim(actions_flatten, inds)))
```
albheim, line 276: Sometimes it is `select_last_dim(to_device(`, and sometimes it is switched like here. I'm sure there is thought behind it but I have not really understood exactly what is going on with all of them.

What would happen if we here (our maybe outside the loop) did `to_device` first and then did the `select_last_dim`? To me it feels like that is what you have tried to do in other places, but I have probably missed something.

findmyway: Technically it is always better to send data `to_device` first. But here the collect trajectory might be too large sometimes. So I guess it's safe to send each sampled batch to the device lazily. It's hard to balance here because the overall performance depends on the size of trajectory, `n_epochs` and `n_microbatches`.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/A2C.jl
findmyway: Because in A2C, only the current state is considered ;)
The actor part calculates the probability of the action taken at the current states, and the critic part calculates the estimated reward of the current states. (If I still remember A2C correctly...)

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/ppo.jl
findmyway: A special note is that the continuous `@view` of a `CuArray` is converted to a `CuArray` directly in the latest version, instead of a `SubArray`. But if it's not continuous, then that might be a problem in the following-up calculations. But if even if a `@view` of `Array` is continuous, when converting to `CuArray`, it will be *adapt*ed and the whole parent of that view is transferred to GPU first, then a `SubArray` of that `CuArray` is returned, instead of a continuous `CuArray`. Maybe we can propose some improvements to `CUDA.jl`.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/A2C.jl
albheim: But shouldn't this pretty much be a noop? In my understanding we have `select_last_dim(S, 1:n) == S`

findmyway: >  it is the length of t but does that mean it is the length of the last dimension of S?

n is the number transitions in a trajectory. For a trajectory of type CircularSARTTrajectory, the (last dimension )length of state and action field is one step longer than the (last dimension ) length of reward and terminal fields. The extra one means next state and action.

findmyway: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/732ecfc2615010189d9c2d402c313351b2591caf/src/ReinforcementLearningCore/src/policies/agents/trajectories/trajectory.jl#L542

albheim: Okay, now I understand. All is good :)

albheim (approved):
LGTM

PR #553, opened 2021-11-29
make bc gpu compatable

findmyway:


findmyway:
@all-contributors please add @andreyzhitnikov for bug

PR #563, opened 2021-12-18
Fix/rand interval

Mo8it:
If at least one of the two endpoints are excluded from an interval, the rand function should not choose it.

---

Why the Union with Array{<:Interval}? I did remove it because it is similar with the functionality in Base:
```julia
julia> rand([ClosedInterval(i, i + 1) for i in 1:10])
7..8

julia> @which rand(fill(ClosedInterval(1.0, 2.0), (2,2)))
rand(X) in Random at /usr/share/julia/stdlib/v1.7/Random/src/Random.jl:259

julia> using ReinforcementLearning

julia> rand([ClosedInterval(i, i + 1) for i in 1:10])
5..6

julia> @which rand(fill(ClosedInterval(1.0, 2.0), (2,2)))
rand(s::Union{Array{<:Interval}, Interval}) in ReinforcementLearningEnvironments at /home/mo/.julia/packages/ReinforcementLearningEnvironments/UhI2A/src/base.jl:9
```

What was the intention of it? I guess, the intention is to get a random action from each interval, not a random interval. This was not working. If this functionality is needed, I can add a commit for it.

---

Also, I did add two methods to choose one of the endpoints if it is not an open interval. This has better performance than rand, but does not restrict to using IntervalSets, because we always have a fallback to rand.

### DIFF
--- a/src/ReinforcementLearningCore/src/policies/agents/agent.jl
+++ b/src/ReinforcementLearningCore/src/policies/agents/agent.jl
@@ -152,6 +152,14 @@ function get_dummy_action(action_space::AbstractVector)
     return action_space[1]
 end
 
+function get_dummy_action(action_space::Interval{:open, :closed})
+    return action_space.right
+end
+
+function get_dummy_action(action_space::Union{Interval{:closed, :open}, Interval{:closed, :closed}})
+    return action_space.left
+end
+
 function RLBase.update!(
     trajectory::AbstractTrajectory,
     policy::AbstractPolicy,
--- a/src/ReinforcementLearningCore/src/policies/agents/agent.jl
+++ b/src/ReinforcementLearningCore/src/policies/agents/agent.jl
@@ -140,6 +140,18 @@ function RLBase.update!(
     push!(trajectory[:terminal], is_terminated(env))
 end
 
+function get_dummy_action(action_space)
+    # For the general case, but especially for continuous action spaces,
+    # we select a random action.
+    # TODO: how to inject a local RNG here to avoid polluting the global RNG
+    return rand(action_space)
+end
+
+function get_dummy_action(action_space::AbstractVector)
+    # For discrete action spaces, we select the first action as dummy action.
+    return action_space[1]
+end
+
 function RLBase.update!(
     trajectory::AbstractTrajectory,
     policy::AbstractPolicy,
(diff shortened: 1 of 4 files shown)


Mo8it:
The first commit is old, sorry. But it is merged anyway.

findmyway:
> What was the intention of it? I guess, the intention is to get a random action from each interval, not a random interval. This was not working. If this functionality is needed, I can add a commit for it.

You guessed it. Since we now have `Space` as the container, we don't need that function any more.

<hr>

Could you resolve the conflict first? So that we can get this merged into the master branch.

Mo8it:
Is it fine now?

src/ReinforcementLearningEnvironments/src/base.jl
```diff
@@ -6,10 +6,18 @@ using MacroTools: @forward
...
 using IntervalSets
 
-Random.rand(s::Union{Interval,Array{<:Interval}}) = rand(Random.GLOBAL_RNG, s)
+Random.rand(s::Interval) = rand(Random.GLOBAL_RNG, s)
 
 function Random.rand(rng::AbstractRNG, s::Interval)
-    rand(rng) * (s.right - s.left) + s.left
+    r = rand(rng)
+    
+    # Check to prevent choosing an exluded endpoint
```
findmyway, line 14:
```suggestion
    # Check to prevent choosing an excluded endpoint
```

src/ReinforcementLearningCore/src/policies/agents/agent.jl
```diff
@@ -152,6 +152,14 @@ function get_dummy_action(action_space::AbstractVector)
...
 end
 
+function get_dummy_action(action_space::Interval{:open,:closed})
+    return action_space.right
+end
+
+function get_dummy_action(action_space::Union{Interval{:closed,:open},Interval{:closed,:closed}})
+    return action_space.left
+end
+
```
findmyway, line 162: This will introduce an extra dependency of `IntervalSets` to `RLCore`. I'm not sure if it is worth adding it here. But if you insist, you'll need to update the https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/master/src/ReinforcementLearningCore/Project.toml (both the `deps` and `compat` sections)

findmyway (changes requested):

src/ReinforcementLearningCore/src/policies/agents/agent.jl
Mo8it: Not really worth it ;)

PR #564, opened 2021-12-21
Remove unneeded method

Mo8it:
Since MonteCarloLearner is an AbstractLearner, this empty method is not needed (default behavior of AbstractLearner).
See here:
https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/8c0a317e35921c7a989597f87f570f58578fb1cf/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/abstract_learner.jl#L22-L27


findmyway:
Thanks!

PR #586, opened 2022-02-25
Default qnetwork initializer

HenriDeh:
Hello,

I'm considering using RL.jl to do some research. I'm currently implementing MPO but when I took a look to the implementation of SAC to get a hang of the inner workings of this package, I noticed some docstrings was incomplete. 
Also, when I tried SAC, I found it cumbersome to specify manually the architecture of the target qnetworks even though they are force copied in the constructor anyways. Finally, the possibility of using a warmup policy is nice but if one does not want to, they still have to specify one. So this PR does three things:

1) It defaults the target networks to copies of the qnetworks 
2) It defaults the start_policy to identity if not used and to `policy` if unspecified by the user
3) It adds some additional information in the docstrings.

I'll soon make some other PRs related to MPO but it is not ready yet.

### DIFF
--- a/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
+++ b/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
@@ -64,9 +64,8 @@ function SACPolicy(;
     policy,
     qnetwork1,
     qnetwork2,
-    target_qnetwork1,
-    target_qnetwork2,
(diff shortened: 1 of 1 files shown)


findmyway:
Nice work. Thanks!

Looking forward to your further contributions! A patch release will be tagged soon.

findmyway:
@all-contributors please add @HenriDeh for code and doc

PR #592, opened 2022-03-04
Fixing and generalizing GaussianNetwork

HenriDeh:
As I was working on a covariance version of the `GaussianNetwork` learner, I noticed a few issues with the latter that this PR fixes.
In +/- the order of the commits:
- RL is not always used for actions in the [-1,1] action-space and thus imposing the `tanh` normalization restrains the applicability of RL.jl to theses applications. This PR adds a field to the `GaussianNetwork` struct that contains the elementwise normalizer function. It is defaulted to `tanh` to avoid breaking changes (if breaking is acceptable, we could default it to `identity` instead). 
- Some algorithms (e.g. MPO) require sampling multiple actions per state. This PR adds a new `GaussianNetwork` caller that takes a batch of states as input and an integer `N` and will sample `N` actions for each state in the batch. As explained in the docstring, we now have to work with 3D tensors for both the state-batch input and the actions output. I made this choice instead of working with a `Vector{Matrix}` input/output mainly for the efficiency of computing the outputs of NN for the entire batch in one pass. 
- While I was at it, Julia asked me to upgrade Manifest.toml to the new format. This is done in this PR, if desired, I can split this to a new one.
- I added a default `GaussianNetwork` initializer for the convenience of not needing to use keywords for the NN (pre, mu, logsigma). Totally optional, I'd understand if you'd rather not have these.
- The most important change, I need someone to confirm me that this was indeed broken and that this is not a misunderstanding on my end. The logpdf returned by the `GaussianNetwork` was broken, this is because it was computed on the unnormalized sampled actions. That is, this test was failing before:
```julia
pre = Dense(20,15)
μ = Dense(15,10)
logσ = Dense(15,10)
gn = GaussianNetwork(pre, μ, logσ)
state = rand(20,3) #batch of 3 states
a, logp = gn(state, is_sampling = true, is_return_log_prob = true)
@test logp == sum(normlogpdf(m, exp.(s), a) .- (2.0f0 .* (log(2.0f0) .- a .- softplus.(-2.0f0 .* a))), dims = 1)
@test logp == gn(state, a)
```
  Now this is fixed.
- I was doing tests for my covariance implementation so I added some for this one too. That's how I found out about the normalization problem above. I think unit tests and documenting should be required for a PR to be merged, otherwise it is simply never added in the future.

That's it. I hope you'll like it and I'm receptive to suggestions and feedback of course.

### DIFF
--- a/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
+++ b/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
@@ -138,6 +138,10 @@ function (model::GaussianNetwork)(state; is_sampling::Bool=false, is_return_log_
     model(Random.GLOBAL_RNG, state; is_sampling=is_sampling, is_return_log_prob=is_return_log_prob)
 end
 
+function (model::GaussianNetwork)(state, action_samples::Int)
+    model(Random.GLOBAL_RNG, state, action_samples)
+end
+
 function (model::GaussianNetwork)(state, action)
     x = model.pre(state)
     μ, raw_logσ = model.μ(x), model.logσ(x) 
--- a/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
+++ b/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
@@ -104,12 +104,12 @@ function (model::GaussianNetwork)(rng::AbstractRNG, state; is_sampling::Bool=fal
         noise = Zygote.ignore() do
             send_to_device(device(model), randn(rng, Float32, size(μ)))
         end
-        z = μ .+ σ .* noise
+        z = model.normalizer.(μ .+ σ .* noise)
         if is_return_log_prob
             logp_π = sum(normlogpdf(μ, σ, z) .- (2.0f0 .* (log(2.0f0) .- z .- softplus.(-2.0f0 .* z))), dims = 1)
-            return model.normalizer.(z), logp_π
+            return z, logp_π
         else
-            return model.normalizer.(z)
+            return z
         end
     else
         return μ, logσ
--- a/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
+++ b/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
@@ -73,7 +73,7 @@ end
 
 Returns `μ` and `logσ` when called.  Create a distribution to sample from using
 `Normal.(μ, exp.(logσ))`. `min_σ` and `max_σ` are used to clip the output from
-`logσ`.
+`logσ`. Actions are normalized according to the specified normalizer function.
 """
 Base.@kwdef struct GaussianNetwork{P,U,S,F}
     pre::P = identity
--- a/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
+++ b/src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
@@ -69,24 +69,25 @@ end
 #####
 
 """
-    GaussianNetwork(;pre=identity, μ, logσ, min_σ=0f0, max_σ=Inf32)
+    GaussianNetwork(;pre=identity, μ, logσ, min_σ=0f0, max_σ=Inf32, normalizer = tanh)
 
 Returns `μ` and `logσ` when called.  Create a distribution to sample from using
 `Normal.(μ, exp.(logσ))`. `min_σ` and `max_σ` are used to clip the output from
 `logσ`.
 """
-Base.@kwdef struct GaussianNetwork{P,U,S}
+Base.@kwdef struct GaussianNetwork{P,U,S,F}
     pre::P = identity
     μ::U
     logσ::S
     min_σ::Float32 = 0f0
     max_σ::Float32 = Inf32
+    normalizer::F = tanh
 end
 
 Flux.@functor GaussianNetwork
 
 """
-This function is compatible with a multidimensional action space. When outputting an action, it uses `tanh` to normalize it.
+This function is compatible with a multidimensional action space. When outputting an action, it uses the `normalizer` function to normalize it elementwise.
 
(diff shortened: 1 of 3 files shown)


src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -101,22 +104,44 @@ function (model::GaussianNetwork)(rng::AbstractRNG, state; is_sampling::Bool=fal
...
+"""
+function (model::GaussianNetwork)(rng::AbstractRNG, state, action_samples::Int)
+    x = model.pre(state)
+    μ, raw_logσ = model.μ(x), model.logσ(x)
+    logσ = clamp.(raw_logσ, log(model.min_σ), log(model.max_σ))
+    
+    σ = exp.(logσ)
+    noise = Zygote.ignore() do
+        send_to_device(device(model), randn(rng, Float32, (size(μ,1), action_samples, size(μ,3))...))
+    end
```
findmyway, line 131: You can simply pass a `RNG` on GPU to generate the noise directly to avoid the overhead of passing data from CPU to GPU.

findmyway (changes requested):

findmyway:
Nice work!

> As I was working on a covariance version of the `GaussianNetwork` learner, I noticed a few issues with the latter that this PR fixes. In +/- the order of the commits:
> 
> * RL is not always used for actions in the [-1,1] action-space and thus imposing the `tanh` normalization restrains the applicability of RL.jl to theses applications. This PR adds a field to the `GaussianNetwork` struct that contains the elementwise normalizer function. It is defaulted to `tanh` to avoid breaking changes (if breaking is acceptable, we could default it to `identity` instead).

Thanks! Better to avoid breaking changes 😃 . And thanks for adding the doc string.

> * Some algorithms (e.g. MPO) require sampling multiple actions per state. This PR adds a new `GaussianNetwork` caller that takes a batch of states as input and an integer `N` and will sample `N` actions for each state in the batch. As explained in the docstring, we now have to work with 3D tensors for both the state-batch input and the actions output. I made this choice instead of working with a `Vector{Matrix}` input/output mainly for the efficiency of computing the outputs of NN for the entire batch in one pass.

👍 

> * While I was at it, Julia asked me to upgrade Manifest.toml to the new format. This is done in this PR, if desired, I can split this to a new one.

It's OK to add it in this PR.

> * I added a default `GaussianNetwork` initializer for the convenience of not needing to use keywords for the NN (pre, mu, logsigma). Totally optional, I'd understand if you'd rather not have these.
> * The most important change, I need someone to confirm me that this was indeed broken and that this is not a misunderstanding on my end. The logpdf returned by the `GaussianNetwork` was broken, this is because it was computed on the unnormalized sampled actions. That is, this test was failing before:
> 
> ```julia
> pre = Dense(20,15)
> μ = Dense(15,10)
> logσ = Dense(15,10)
> gn = GaussianNetwork(pre, μ, logσ)
> state = rand(20,3) #batch of 3 states
> a, logp = gn(state, is_sampling = true, is_return_log_prob = true)
> @test logp == sum(normlogpdf(m, exp.(s), a) .- (2.0f0 .* (log(2.0f0) .- a .- softplus.(-2.0f0 .* a))), dims = 1)
> @test logp == gn(state, a)
> ```
> 

What are `m` and `s` here? I'd use `isapprox` for comparison here.

> Now this is fixed.
> 
> * I was doing tests for my covariance implementation so I added some for this one too. That's how I found out about the normalization problem above. I think unit tests and documenting should be required for a PR to be merged, otherwise it is simply never added in the future.
> 
> That's it. I hope you'll like it and I'm receptive to suggestions and feedback of course.

Thanks for adding those tests. See inline comment to avoid breaking existing CI pipeline.

HenriDeh:
Sorry I forgot to copy a line. 
```
pre = Dense(20,15)
μ = Dense(15,10)
logσ = Dense(15,10)
gn = GaussianNetwork(pre, μ, logσ)
state = rand(20,3) #batch of 3 states
m, s = gn(state)
a, logp = gn(state, is_sampling = true, is_return_log_prob = true)
@test logp == sum(normlogpdf(m, exp.(s), a) .- (2.0f0 .* (log(2.0f0) .- a .- softplus.(-2.0f0 .* a))), dims = 1)
@test logp == gn(state, a)`
```
I'll change to isapprox yes, oops.

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
HenriDeh: Yes indeed, here I copied the practice from the current implementation (see line 105 of the same file). But basically, since rng is passed by the algorithm and is decided by the user, we can just do 
```suggestion
    noise = Zygote.ignore() do
        randn(rng, Float32, (size(μ,1), action_samples, size(μ,3))...)
    end
```
and assume the user passed a rng on the same device as the GN?

findmyway: Yep

HenriDeh:
I changed the rand of other NN approximators then to be consistent. I think this will break CI though, we'll have to pass an rng explicitly when testing on the GPU.

findmyway:
Yes, you can add a device rng like this:

https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/f4cf555f50e0333254ab099f5a643a587ec532d7/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_IQN_CartPole.jl#L25

HenriDeh:
> Yes, you can add a device rng like this:
> 
> https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/f4cf555f50e0333254ab099f5a643a587ec532d7/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_IQN_CartPole.jl#L25

Since the test is already in a `if CUDA.functional()` block, I directly added the CUDA rng as argument.

I'm a bit worried: the test were successful before I commited that, does it mean that they do not run the CUDA.functional() block ?

HenriDeh:
Commit [f7348da](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/592/commits/f7348da2a2d097941397345e3bc6672065aeebbe) makes the PR breaking actually. I don't see a way to remove the device transfer overhead without breaking existing implementations though :/

PR #597, opened 2022-03-10
Add CovGaussianNetwork to work with covariance

HenriDeh:
This PR adds a new approximator that learns a full covariance matrix instead of the diagonal (i.e. GaussianNetwork).
This is essentially a building block for the MPO algorithm that I intend to implement next. 
I implemented this with performance in mind, notably I checked that it works with CUDA, but it's totally possible that I missed some opportunities for improvements. 

The CGN works as follows:
- It takes a batch of states as input. The batch must be 3D with the last dim being the batch dim because Covariances are matrices not vectors anymore.
- It returns a (batch of) mean vectors and `cholesky_vector`. That cholesky vector contains the non-zero elements of the lower-triangular Cholesky decomposition of a covariance matrix. We then reshape the vector into a lower triangular matrix `L` by adding the zeros at the right places. We also need to softplus the diagonal elements to ensure positivity. That is, the actual covariance matrix is L*L'. We don't actually ever need to use the explicit covariance, it's more efficient to directly work with `L`. 

And that's it really. The difficulty was mainly the reconstruction of L in a gpu and AD friendly manner (no mutation, scalar indexation, etc.)


HenriDeh:
Only the 5 last commits are new stuff, the rest is from my previous PR. I'm trying to see how I can remove that.

HenriDeh:
Side note: since CGN works with 3D tensors, I don't think it is compatible out of the box with PPO, SAC, etc. But I think it would be straightforward to do that using some function overloading.

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -151,6 +153,160 @@ function (model::GaussianNetwork)(state, action)
...
+    else
+        Σ = Flux.stack(map(l -> l*l', eachslice(L, dims=3)),3)
+        return μ, Σ
+    end
+end
+
+"""
+    (model::CovGaussianNetwork)(rng::AbstractRNG, state::AbstractMatrix; is_sampling::Bool=false, is_return_log_prob::Bool=false)
+    
+    Given a Matrix of states, will return actions, μ and logpdf in matrix format. The batch of Σ remains a 3D tensor.
```
findmyway, line 215:
```suggestion
Given a Matrix of states, will return actions, μ and logpdf in matrix format. The batch of Σ remains a 3D tensor.
```

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -151,6 +153,160 @@ function (model::GaussianNetwork)(state, action)
...
+    logp_π = mvnormlogpdf(μ, L, z)
+    return z, logp_π
+end
+
+function (model::CovGaussianNetwork)(state::AbstractArray, args...; kwargs...)
+    model(Random.GLOBAL_RNG, state, args...; kwargs...)
+end
+
+"""
+    (model::CovGaussianNetwork)(state, action)
```
findmyway, line 256:
```suggestion
    (model::CovGaussianNetwork)(state, action)
    
```

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -151,6 +153,160 @@ function (model::GaussianNetwork)(state, action)
...
+        tc_other = cholesky_vec[c2idx(j,j)+1:c2idx(j+1,j+1)-1,:,:]
+        zs = Flux.Zygote.ignore() do 
+            zeros(eltype(tc_diag), da - size(tc_other,1) - 1,1,batch_size)
+        end
+        [zs; tc_diag; tc_other]
+    end
+    return mapreduce(f, hcat, 1:da)
+end
+
+function trilcol(cholesky_vec::CuArray,da)
```
findmyway, line 296: Is it possible to remove the constraint of `CuArray` here?

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
```diff
@@ -151,6 +153,160 @@ function (model::GaussianNetwork)(state, action)
...
+    batch_size = size(cholesky_vec, 3)
+    c2idx(i,j) = ((2da-j)*(j-1))÷2+i #return the position in cholesky_vec of the element of the triangular matrix at coordinates (i,j)
+    function f(j) #return a slice (da x 1 x batchsize) containing the jth columns of the lower triangular cholesky decomposition of the covariance
+        tc_diag = softplus.(cholesky_vec[c2idx(j,j):c2idx(j,j),:,:])
+        tc_other = cholesky_vec[c2idx(j,j)+1:c2idx(j+1,j+1)-1,:,:]
+        zs = Flux.Zygote.ignore() do 
+            CUDA.zeros(eltype(tc_diag), da - size(tc_other,1) - 1,1,batch_size)
```
findmyway, line 303: I think we can use `similar` to create an uninitialized array and then fill it with zero. So that we can avoid using `CUDA` explicitly here.

findmyway (approved):

findmyway:
LGTM

@albheim might also be interested in this PR 😄

albheim:
Seems like a good thing to have, sadly a bit too swamped right now to have a proper look.

src/ReinforcementLearningCore/src/policies/q_based_policies/learners/approximators/neural_network_approximator.jl
HenriDeh: Yes indeed, that was my first approach but backprop did not work because that's an array mutation. Only later I had to ignore CUDA.zeros as well and I did not think that I could have done that with the original implementation. SO yes, I'll make that change thanks!

HenriDeh:
Oh, what do you think about the name `CovGaussianNetwork` ? I chose it when I started without giving it more than a though, maybe another one would be more appropriate. Other possibilities are notably:
- FullGaussianNetwork (used internally in [DistributionsAD.jl](https://github.com/TuringLang/DistributionsAD.jl) to differenciate from Diag and Scalar Gaussians)
- MvGaussianNetwork

HenriDeh:
Btw: 
image
`CUDA.functional()` is false in the CI jobs.

findmyway:
> Oh, what do you think about the name `CovGaussianNetwork` ? I chose it when I started without giving it more than a though, maybe another one would be more appropriate. Other possibilities are notably:

I think that's fine. We can rename it later if someone else has a strong objection.

src/ReinforcementLearningCore/test/extensions.jl
```diff
@@ -17,23 +17,78 @@ end
...
+            CUDA.@allowscalar @test (mapslices(tril!, g_d[L_d], dims=(1,2)) |> Array) ≈ mapslices(tril!, g[L], dims=(1,2))
         end
-        @test isapprox(cpu_grad[1], gpu_grad[1] |> Array)
+    end
+end
+
+@testset "logdetLorU" begin
+    L = tril(sqrt.(rand(Float32,5,5) .^2))
+    Σ = L*L'
+    @test logdet(Σ) ≈ RLCore.logdetLorU(L)
```
findmyway, line 90: I just noticed that this line failed occasionally.

See example: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/runs/5580190385?check_suite_focus=true#step:10:531

HenriDeh: I take care of this in #603

PR #600, opened 2022-03-13
Fixes CircularArrayBuffers dependency

harwiltz:
I encountered a dependency issue when trying to run some experiments that appears to be caused by an update to the `CircularArrayBuffers` package. My project environment had `CircularArrayBuffers@0.1.7` installed and was causing the following error when I tried to run the `JuliaRL_IQN_CartPole` experiment (among others):

```
ERROR: MethodError: no method matching _buffer_frame(::CircularArrayBuffers.CircularVectorBuffer{Float32, Vector{Float32}}, ::Matrix{Int64})

You might have used a 2d row vector where a 1d column vector was required.
Note the difference between 1d column vector [1,2,3] and 2d row vector [1 2 3].
You can convert to a column vector with the vec() function.
Closest candidates are:
  _buffer_frame(::CircularArrayBuffers.CircularArrayBuffer, ::Int64) at ~/.julia/packages/CircularArrayBuffers/FWC6A/src/CircularArrayBuffers.jl:70
  _buffer_frame(::CircularArrayBuffers.CircularArrayBuffer, ::AbstractVector{<:Integer}) at ~/.julia/packages/CircularArrayBuffers/FWC6A/src/CircularArrayBuffers.jl:80
```

The problem resolves itself as soon as I downgrade `CircularArrayBuffers`.

This is probably more like a band-aid than a permanent fix, its probably just a simple change somewhere. But I'm not very familiar with `CircularArrayBuffers` so I thought this fix could be helpful in the meantime.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,6 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
+julia_version = "1.7.1"
 manifest_format = "2.0"
+
+[[AbstractFFTs]]
+deps = ["ChainRulesCore", "LinearAlgebra"]
+git-tree-sha1 = "6f1d9bc1c08f9f4a8fa92e3ea3cb50153a1b40d4"
+uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
+version = "1.1.0"
 
 [[deps.AbstractFFTs]]
 deps = ["ChainRulesCore", "LinearAlgebra"]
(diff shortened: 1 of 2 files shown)


findmyway:
Thanks for reporting. Should be fixed in https://github.com/JuliaReinforcementLearning/CircularArrayBuffers.jl/pull/11

Please update the dependency to the `v0.1.8` once https://github.com/JuliaRegistries/General/pull/56499 get merged.

findmyway:
@all-contributors please add @harwiltz for bug

harwiltz:
@findmyway great, will do. Thanks!

PR #604, opened 2022-03-17
WIP: Add MPO in zoo

HenriDeh:
I'm opening this as a draft so discussions are possible early. 
This implements the MPO algorithm from [this paper](https://arxiv.org/abs/1806.06920) and [its improved version](https://arxiv.org/abs/1812.02256)
PR Checklist

- [x] Add docstrings
- [x] Handle the case of a discrete actor. For this, I was wondering if a `DiscreteNetwork` akin to `GaussianNetwork` may be a better approach than considering that if a GN is not used, then it must be a Discrete actor. 
- [x] Add some tests
- [x] Does this handle distributed environments?
- [x] Handle legal action masks
- [x]  Decide default HPs
- [x]  Make experiments with each network
- [x]  Remove normalizer from networks ?
- [x]  Make a dedicated doc page 
- [x]  Fix GPU (changes in RLTrajectories needed)


findmyway:
Nice work!

I may not have the time to review it in the next two or three weeks. This PR is relatively independent, so feel free to merge it if you think it's ready.

findmyway:
> Maybe use a more general batch sampling. 

It's on my todo list but I never got the time to implement it. Ideally the sampling step is independent while the training process is reactive. This may introduce some breaking changes to the current design. I'll write down my thoughts and discuss with you later.

PR #621, opened 2022-04-20
Episode reset condition

HenriDeh:
This PR adds a reset_condition argument to the `run` method and built-in conditions. It is defaulted to `ResetAtTerminal()` to avoid breaking changes, which will trigger the end of the episode loop when is_terminated is true.

The new possibility is to use `ResetAfterNSteps(n)` instead. I made a doc page for more information.   

- [ ] Update NEWS.md?
- [x] Before this PR can be merged, we must wait to see the final design of Trajectories.jl. A priori, it should easily integrate. The main point to be careful about is that dummy states should no longer exist, as planned in the new design.


HenriDeh:
As expected, this did not require many changes to work. Are you okay with this addition @findmyway  ?

findmyway:
LGTM

PR #625, opened 2022-04-26
Add a categorical Network

HenriDeh:
Hi,

I made a categorical network to work with discrete action spaces using the same api than GaussianNetworks. 
However I'm a bit stuck: how do we sample on the GPU for a categorical distribution? We must use the provided rng to keep replicability, but that means we cannot move the probability parameters to the cpu to perform the sampling if rng is from CUDA. 

Also, I need to add the possibility of action masks, but I don't know how that works yet.

PR Checklist

- [ ] Update NEWS.md?
- [x] Action masks
- [x] Gpu sampling


HenriDeh:
Thank you, that should work I believe. The issue was that using `rand(CUDA.RNG(), Categorical(cu([0.1,0.2,0.7])))` uses scalar indexing. I tried and `argmax(cu(rand(5)))` does not use scalar indexing so the Gumbel trick should do it.

HenriDeh:
I'm also resuming this PR. I used the Gumbel-Max trick described in the paper you linked to enable gpu sampling (cool stuff thanks for the reference). I have some points to discuss before merging this:
- Unlike (Cov)GaussianNetwork, I went for `is_return_logits` instead of `is_return_log_prob`. The reason is that it is more usual to work with logits in loss functions when dealing with categorical distributions. Do you agree that it is the best choice or do you feel that log_probs would be better? It would not impact stability to do the latter, it's just a convention choice.
- I made tests but they are not included yet, I'm waiting for the manifest shenanigans to be solved before moving on.
- I found out that the tests for the other networks (Gaussian etc.) were no longer present. I don't know if this is intentional or not. I added them back for now, though some are commented because they were reworked recently.
- I did not add a `(model)(state, action)` method, unlike gaussians. Should I? I don't know if it makes sense because, again, we work with logits.

HenriDeh:
Hello @findmyway, do you have any suggestion/comment/review on this Categorical Network ? Otherwise I'll merge after CI passes.

src/ReinforcementLearningCore/src/utils/networks.jl
```diff
@@ -271,6 +273,122 @@ function vec_to_tril(cholesky_vec, da)
...
+- `is_return_logits::Bool=false`, whether to return the *logits* of getting the sampled actions in the given state.
+Only applies if `is_sampling` is true and will return `z, logits`.
+
+If `is_sampling = false`, returns only the logits obtained by a simple forward pass into `model`.
+"""
+mutable struct CategoricalNetwork{P}
+    model::P
+end
+
+Flux.@functor CategoricalNetwork
```
findmyway, line 302:
```suggestion
@functor CategoricalNetwork
```

findmyway (approved):
LGTM

Issue #614, opened 2022-04-04
Next Release Plan (v0.11)


findmyway:
# Goal

Improve the interactions between ReinforcementLearning.jl and other ecosystems in Julia.

## Why is it important?

In the early days of developing this package, the main goal is to reproduce some popular (deep) RL algorithms. It's still important to keep adding new emerging algorithms into this package. But as an engineer, I always think the higher impact is achieved only when users really **apply** those powerful RL algorithms to the problems they are interested in. In recent years, many important packages across different domains were developed in Julia and the whole ecosystem improved a lot. Although the interfaces defined in this package are loose and flexible, people are still unsure how to use this package due to lacking concrete examples. Adding more examples and removing some restricted assumptions will greatly encourage more people to try this package. On the other hand, doing so will also improve the quality of this package.

## Potential breaking changes

The most important change would be decoupling the training data generation and policy optimization. The state is assumed to be a tensor by default in many cases. This is the main blocking issue when interacting with many other packages. Besides, the async training pipeline will not only improve the performance of existing algorithms on a single node but also provide the foundation of large scale training in future releases (possibly in v0.12)

## Key issues to be addressed

Following are some of the existing issues on the top of my mind. Please raise new ones if you wish to be addressed in the next release.

### Environments

- [x] https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/268
- [x] https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/572 ~~Stil no luck to address this issue, so I have to remove OpenSpiel related part in the next release.~~ https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/691 
- [ ] Demonstrate that environments can be implemented on not only CPU but also GPU   https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/121
- [ ] https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/546
- [ ] Take a look at https://github.com/dojo-sim/Dojo.jl
- [ ] Take a look at https://github.com/Lyceum/MuJoCo.jl 
- [ ] More examples with DifferentialEquations.jl
- [ ] https://github.com/deepmind/android_env Should be OK through PyCall.jl. Need some examples.
- [ ] Look into https://github.com/corail-research/SeaPearl.jl and see if there could be some improvements.

### Refactor Existing Policies

- [x] BasicDQN
[...]

PR #632, opened 2022-05-08
Use Trajectories.jl instead

findmyway:
This PR aims to split out trajectories related codes and simplify the code logic in RLCore


findmyway:
Should be added back soon. This package is slowly moving towards v0.11 You may track the progress here: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/614

findmyway:
> what's the reason to move away from `::AbstractLearner` and `::AbstractApproximator`?

AbstractLearner was removed by accident. (And is already added back, it's just that the rename part is not reverted yet)

I find the old NeuralNetworkApproximator is general enough. So simply removed the abstract type and renamed it to Approximator

Issue #648, opened 2022-06-06
`TDLearner` time step parameter


baedan:
`TDLearner(;approximator, γ=1.0, method, n=0)`: the `n` in the constructor is strangely not the number of time steps used, but rather that number minus 1.  this is really strange.

findmyway:
;(
I struggled on it too...
In the end, I decided to follow `TD(λ)` where (`λ=0`). So maybe better to rename the keyword argument name?

findmyway:
Hmm, let me examine it again when adding it back in the next release.

PR #649, opened 2022-06-06
added basic doc for `TDLearner`

baedan:
PR Checklist

- [ ] Update NEWS.md?


baedan:
pending #648

src/ReinforcementLearningZoo/src/algorithms/tabular/td_learner.jl
```diff
@@ -3,6 +3,18 @@ export TDLearner
 using LinearAlgebra: dot
 using Distributions: pdf
 
+"""
+    TDLearner(;approximator, γ=1.0, method, n=0)
+
+Use temporal-difference method to estimate state value or state-action value.
+
+# Fields
+- `approximator` can be either
```
HenriDeh, line 12:
```suggestion
- `approximator` can be either a
```
To emphasize that what comes after is a list of types, not a list of accepted values (to the difference with `method` below).

baedan:
this can merge now, with the caveat that it might need an update pending #648

PR #726, opened 2022-07-07
Adapted SAC to support MultiThreadedEnv

BigFood2307:
Hey everyone!

I was recently trying to use SAC with the MultiThreadedEnv Wrapper and found that I needed to adapt the Code a little for that to work.
I decided to share my version of sac.jl that now works with "Normal" Envs and the MTE as well.

I will be honest: I am not very experienced with Julia and may have missed some stuff. Also I was not able to properly test the Code inside the ReinforcementLearning.jl repo. It works perfectly when I define a seperate MSACPolicy with these changes in my local project however.
I hope someone can double check my stuff and possibly give me some hints on how to test it myself for later reference.

Also here would be the Code for a Experiment using a MTE (If I havent missed anything):

```
function RL.Experiment(
    ::Val{:JuliaRL},
    ::Val{:SAC},
    ::Val{:Pendulum},
    ::Nothing;
    save_dir=nothing,
    seed=123
)
    rng = StableRNG(seed)
    inner_env = PendulumEnv(T=Float32, rng=rng)
    action_dims = inner_env.n_actions
    A = action_space(inner_env)
    low = A.left
    high = A.right
    ns = length(state(inner_env))
    na = 1
    env = MultiThreadEnv[ActionTransformedEnv(
        PendulumEnv(T=Float32, rng=StableRNG(seed+i));
        action_mapping=x -> low + (x[1] + 1) * 0.5 * (high - low)
    ) for i in 1:N_ENV])
    init = glorot_uniform(rng)

    create_policy_net() = NeuralNetworkApproximator(
        model=GaussianNetwork(
            pre=Chain(
                Dense(ns, 30, relu, init=init),
                Dense(30, 30, relu, init=init),
            ),
            μ=Chain(Dense(30, na, init=init)),
            logσ=Chain(Dense(30, na, x -> clamp(x, typeof(x)(-10), typeof(x)(2)), init=init)),
        ),
        optimizer=ADAM(0.003),
    ) |> gpu

    create_q_net() = NeuralNetworkApproximator(
        model=Chain(
            Dense(ns + na, 30, relu; init=init),
            Dense(30, 30, relu; init=init),
            Dense(30, 1; init=init),
        ),
        optimizer=ADAM(0.003),
    ) |> gpu

    agent = Agent(
        policy=SACPolicy(
            policy=create_policy_net(),
            qnetwork1=create_q_net(),
            qnetwork2=create_q_net(),
            target_qnetwork1=create_q_net(),
            target_qnetwork2=create_q_net(),
            γ=0.99f0,
            τ=0.005f0,
            α=0.2f0,
            batch_size=64,
            start_steps=1000,
            start_policy=RandomPolicy(Space([Space([-1.0..1.0 for _ in 1:na]) for _ in 1:N_ENV]); rng=rng),
            update_after=1000,
            update_freq=1,
            automatic_entropy_tuning=true,
            lr_alpha=0.003f0,
            action_dims=action_dims,
            rng=rng,
            device_rng=CUDA.functional() ? CUDA.CURAND.RNG() : rng
        ),
        trajectory=CircularArraySARTTrajectory(
            capacity=10000,
            state = Matrix{Float32} => (ns, N_ENV),
            action = Matrix{Float32} => (na, N_ENV),
            reward = Vector{Float32} => (N_ENV,),
            terminal = Vector{Bool} => (N_ENV,),
        ),
    )

    stop_condition = StopAfterStep(10_000, is_show_progress=!haskey(ENV, "CI"))
    hook = TotalBatchRewardPerEpisode()
    Experiment(agent, env, stop_condition, hook, "# Play Pendulum with SAC")
end
```

I hope this is helpfull :D

### DIFF
--- a/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
+++ b/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
@@ -119,13 +119,18 @@ function (p::SACPolicy)(env)
     p.update_step += 1
 
     if p.update_step <= p.start_steps
-        p.start_policy(env)
+        action = p.start_policy(env)
+        if(size(action[1]) != ())
+            action = permutedims(hcat(action...))
+        end
+        action
     else
         D = device(p.policy)
         s = send_to_device(D, state(env))
         s = Flux.unsqueeze(s, dims=ndims(s) + 1)
         # trainmode:
-        action = dropdims(p.policy(p.device_rng, s; is_sampling=true), dims=2) # Single action vec, drop second dim
+        action = p.policy(p.device_rng, s; is_sampling=true)
+        action = dropdims(action, dims=ndims(action)) # Single action vec, drop second dim
         send_to_host(action)
 
         # testmode:
@@ -156,18 +161,18 @@ function RLBase.update!(p::SACPolicy, batch::NamedTuple{SARTS})
     q′_input = vcat(s′, a′)
     q′ = min.(p.target_qnetwork1(q′_input), p.target_qnetwork2(q′_input))
 
-    y = r .+ γ .* (1 .- t) .* vec(q′ .- α .* log_π)
+    y = r .+ γ .* (1 .- t) .* dropdims(q′ .- α .* log_π, dims=1)
 
     # Train Q Networks
     q_input = vcat(s, a)
 
     q_grad_1 = gradient(Flux.params(p.qnetwork1)) do
-        q1 = p.qnetwork1(q_input) |> vec
+        q1 = dropdims(p.qnetwork1(q_input), dims=1)
         mse(q1, y)
     end
     update!(p.qnetwork1, q_grad_1)
     q_grad_2 = gradient(Flux.params(p.qnetwork2)) do
-        q2 = p.qnetwork2(q_input) |> vec
+        q2 = dropdims(p.qnetwork2(q_input), dims=1)
         mse(q2, y)
     end
     update!(p.qnetwork2, q_grad_2)
--- a/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
+++ b/src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
@@ -121,7 +121,7 @@ function (p::SACPolicy)(env)
     if p.update_step <= p.start_steps
         action = p.start_policy(env)
         if(size(action[1]) != ())
-            action = permutedims(hcat(action...))
+            action = reduce(hcat, action)
         end
         action
     else


findmyway:
Hi @BigFood2307 , 

This is helpful! The change looks fine to me. This PR will be merged later when adapting the SAC to the new interfaces in the master branch.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/sac.jl
```diff
@@ -119,13 +119,18 @@ function (p::SACPolicy)(env)
     p.update_step += 1
 
     if p.update_step <= p.start_steps
-        p.start_policy(env)
+        action = p.start_policy(env)
+        if(size(action[1]) != ())
+            action = permutedims(hcat(action...))
```
HenriDeh, line 124:
```suggestion
            action = permutedims(reduce(hcat, action))
```
In Julia (you mentioned you are new to the language) you want to avoid using the splat operator to do this kind of stuff because it's really slow. `reduce(hcat, yourvectorofvectors)` does that much faster.

I also want to ask, why are the dimensions permuted afterward? If `action` is a collection of vectors, your action matrix will have actions as rows instead of columns. If you come from python you should know that Julia matrices are "column-major". So if this is why you used permutedims, it should also be removed.

BigFood2307: I found that I had to reshape the output of the RandomPolicy, since it gives me a vector of vectors instead of a matrix. Thinking about it, this might indeed be an issue if one wants to use any other start_policy.
There might just be a better way to generate the random actions in the beginning for multiple envs. I guess the issue is, that a RandomPolicy will not adjust if it is working on a MultiThreadedEnv and my solution of using a Space(Space(...)) in the RP will give me those Vectors. Do you maybe have a suggestion on how to setup the RandomPolicy for MutliThreadedEnvs? That would basically make the whole if statement unnecessary.

HenriDeh: Hi,

I'm not sure I understand your message. Basically what I mean regarding permutedims is this : 
image
The actions should be the columns of the resulting matrix, whereas permutedims leads to row actions. I think the if statement is fine. I haven't dealt with MultiThreadedEnvs enough to investigate this right now but if this remains a problem later I can check it out

HenriDeh: Can you perhaps show me a screenshot of this problem?

BigFood2307: I started setting up an example when I found an easy solution:
I just had to switch around the 2 Spaces inside the RandomPolicy definition (basically na <--> N_ENV). Then i dont need the permutedims. Not sure why I never tried that. Its more intuitive to me this way around anyways, so obviously the better solution.
I updated the experiment in my first comment and pushed the better version to my branch.
Thanks for helping me understand how that stuff works.

HenriDeh: Perfect!

PR #733, opened 2022-07-17
add VPG

findmyway:
wait https://github.com/JuliaRegistries/General/pull/64390


src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vpg.jl
```diff
@@ -1,147 +1,87 @@
...
+    end
+
+    gs = gradient(params(A)) do
+        if p.dist <: DiscreteDistribution
+            log_prob = s |> A |> logsoftmax
+            log_probₐ = log_prob[CartesianIndex.(a, 1:length(a))]
+        elseif p.dist <: ContinuousDistribution
+            dist = p.dist.(A(s)...) # TODO: this part does not work on GPU. See: https://github.com/JuliaStats/Distributions.jl/issues/1183 .
+            log_probₐ = logpdf.(dist, A)
+        end
```
baedan, line 79: i'm not sure this would work.  `dist` maps the output of the network to a distribution, so it's a function type in general.  in the discrete case, where `dist = Categorical`, `Categorical` is coincidentally the type as well as the constructor, so `p.dist <: DiscreteDistribution` would match.  but in general, `dist` need not be a `Distribution` type.

findmyway: Yes, here `dist` is a mapping from `parameters... -> distribution`. A more conventional way is to add a custom trait to specify the distribution type of the result. But given that most distributions provided in `Distributions.jl` are both type and function, I think it's safe to assume `dist` is always a struct type here.

baedan: oh!  the example i had in mind was `Normal(μ,σ)` which has two arguments but i guess that works too with `A(s)...`, nice

PR #737, opened 2022-07-19
implement `action_distribution`

baedan:

PR #747, opened 2022-08-08
TRPO

baedan:
this PR implements Trust-Region Policy Optimization, and adds a CartPole experiment for it.

to this end, i wrote a few utility functions that are shared amongst policy gradient policies (#737).  but perhaps a better way to go about it is to have a `PolicyGradientPolicy` type, and have it wrap different learners.


findmyway:
Looks fine to me in general. I think there's still room for improvement in the `gradient` part. I'll add more detailed comments this weekend.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/util.jl
```diff
@@ -0,0 +1,181 @@
...
+5-element Vector{Normal{Float64}}:
+ Normal{Float64}(μ=1.0, σ=2.0)
+ Normal{Float64}(μ=3.0, σ=4.0)
+ Normal{Float64}(μ=5.0, σ=6.0)
+ Normal{Float64}(μ=7.0, σ=8.0)
+ Normal{Float64}(μ=9.0, σ=10.0)
+```
+"""
+action_distribution(dist::Type{T}, model_output) where {T<:ContinuousDistribution} = 
+    map(col -> dist(col...), eachcol(model_output))
```
findmyway, line 69: This has some extra assumptions here. 

1. Parameters of the distribution are of the same length and size (scalar to be more specific)
2. The output of the model is a `Matrix`

Maybe we can figure out a more elegant way to use `StructArrays.jl` here later.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/util.jl
```diff
@@ -0,0 +1,181 @@
...
+
+"""
+    surrogate_advantage(model, states, actions, advantage, action_logits)
+Calculate the surrogate advantage of `model` on experiences (`states` and `actions`)
+generated by an old model, which generated `action_logits`. 
+
+See [here](https://spinningup.openai.com/en/latest/algorithms/trpo.html#key-equations) for more information.
+"""
+function surrogate_advantage(model, states, actions, advantage, action_logits)
+    π_θₖ = map(eachcol(softmax(action_logits)), actions) do a, b
```
findmyway, line 144: Same as above

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/util.jl
```diff
@@ -0,0 +1,181 @@
...
+        a[b]
+    end
+    π_θ = map(eachcol(softmax(model(states))), actions) do a, b
```
findmyway, line 147: Same as above

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vpg.jl
```diff
@@ -54,9 +49,12 @@ function RLBase.optimise!(p::VPG, batch::NamedTuple{(:state, :action, :gain)})
     A = p.approximator
     B = p.baseline
     s, a, g = map(Array, batch) # !!! FIXME
+    local δ
 
     if isnothing(B)
         δ = normalise(g)
+        loss = 0
+        println("branch")
```
findmyway, line 57:
```suggestion
```

findmyway (approved):

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/trpo.jl
```diff
@@ -0,0 +1,113 @@
...
+            ignore_derivatives() do
+                # @info "TRPO/baseline" loss = loss δ
+            end
+            loss
+        end
+        optimise!(B, gs)
+    end
+    
+    # store logits as intermediate value
+    old_logits = Ref{Matrix{Float32}}()
```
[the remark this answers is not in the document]
baedan: there were some oddities with Zygote 2nd order derivatives wrt implicit parameters when i tried `local old_logits`, yielding inconsistent results after `mapreduce(vec, vcat, gradient)` (i got a `Ref(0)` term sometimes as the first term).  i'm not sure this is necessary anymore however, since i've since changed how the 2nd order gradient is calculated.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/util.jl
```diff
@@ -0,0 +1,181 @@
+using Distributions: DiscreteDistribution, ContinuousDistribution
+using Flux: softmax, gradient, params
+using NNlib: logsoftmax
+
+export action_distribution, policy_gradient_estimate, IsPolicyGradient
+export conjugate_gradient!
+
+struct IsPolicyGradient end
```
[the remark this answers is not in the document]
baedan: i think so!  or better yet, a `PolicyGradient` wrapper class perhaps?

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/util.jl
baedan: yeah, the semantics of `dist` is pretty bad here.  after [this](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/733#discussion_r923176416) i realized punning `dist` wouldn't even work with a `Normal`, since we usually want the network to output the log of the variance, not the variance itself.

perhaps here we could just ask the user to specify distribution type with a trait, and overload a `dist` function (with a better name of course).

findmyway:
I'll merge this first. I may find some time in the next week to polish this further ;)

PR #750, opened 2022-08-21
Update DoEveryNEpisode hook to new api

ludvigk:
DoEveryNEpisode hook was called with stage=POST_EPISODE_STAGE as default rather than PostEpisodeStage().

Not sure how this is meant to work in the future, but this is at least a temporary fix.


findmyway:
Thanks!

@all-contributors please add @ludvigk for code

PR #776, opened 2022-12-16
Trying to reimplement experiments

HenriDeh:
I'm trying to uncomment the experiments. I cannot do it locally without having to `dev` all packages, which is extremely inconvenient. I think it may work in github CI's though.


HenriDeh:
Some experiments were broken due to the action_space function of Cartpole missing the player argument.

PR #827, opened 2023-03-07
Basic allocation fixes for RandomWalk / RandomPolicy

jeremiahpslewis:
PR Checklist

- [x] Update NEWS.md?

```julia
# JET: 117 possible errors -> 0
# BenchmarkTools on
using ReinforcementLearningBase
using ReinforcementLearningCore
using ReinforcementLearningEnvironments
using BenchmarkTools
using JET

env = RandomWalk1D()
policy = RandomPolicy(legal_action_space(env))
@btime run(
    policy,
    env,
    StopAfterEpisode(1_000_000; is_show_progress=false),
    TotalRewardPerEpisode(; is_display_on_exit=false),
)
# Previously: 1.040 s (8991309 allocations: 146.98 MiB)
# This PR: 66.632 ms (16 allocations: 9.78 MiB)
# This PR, with `is_show_progress=true`, `is_display_on_exit=true`: 323.874 ms (4991 allocations: 26.33 MiB)
```


src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -104,32 +104,44 @@ Base.getindex(h::RewardsPerEpisode) = h.rewards
...
+    function TotalRewardPerEpisode(; is_display_on_exit::Bool = true)
+        struct_type = is_display_on_exit ? Bool : Nothing
+        new{struct_type}(Float64[], 0.0, is_display_on_exit)
+    end
 end
 
 Base.getindex(h::TotalRewardPerEpisode) = h.rewards
 
-(h::TotalRewardPerEpisode)(::PostActStage, agent, env) = h.reward += reward(env)
+(h::TotalRewardPerEpisode)(::PostActStage, agent::AbstractPolicy, env::AbstractEnv) =
```
HenriDeh, line 120: Is this necessary for type stability ? Looks like this only restricts the type of agent and env.

src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -104,32 +104,44 @@ Base.getindex(h::RewardsPerEpisode) = h.rewards
 Store the total reward of each episode in the field of `rewards`. If
 `is_display_on_exit` is set to `true`, a unicode plot will be shown at the [`PostExperimentStage`](@ref).
 """
-Base.@kwdef mutable struct TotalRewardPerEpisode <: AbstractHook
-    rewards::Vector{Float64} = Float64[]
-    reward::Float64 = 0.0
-    is_display_on_exit::Bool = true
+mutable struct TotalRewardPerEpisode{T} <: AbstractHook where {T<:Union{Bool,Nothing}}
```
HenriDeh, line 107: Is this a type stability change ?

HenriDeh: I have the same question for most changes below in fact. I guess the point is to use multiple dispatch to compile different functions (one that plots and one that doesn't)? Does this change significantly impact performance ?

jeremiahpslewis: The plot and ProgressMeter functions seems to be type unstable, so if you don't need them, it's better if they aren't run. As a second thing, the plot / progress meter status was being checked at every hook / stop call, even though this is already known at type creation. So two reasons why it's useful; admittedly the Bool/Nothing in `hooks.jl` is a bit weird, but it follows the existing pattern in `stop.jl` with the progress meter

src/ReinforcementLearningCore/src/core/hooks.jl
jeremiahpslewis: I'll drop this from this PR. But I think in a separate change, we should only define behavior on `agent::AbstractPolicy` and `env::AbstractEnv`; this may have performance relevance (I don't believe it does in this case), but from a user experience perspective it's helpful to see it in the code / function signature and should error if random other objects are used.

src/ReinforcementLearningCore/Project.toml
```diff
@@ -36,7 +36,7 @@ FillArrays = "0.8, 0.9, 0.10, 0.11, 0.12, 0.13"
 Flux = "0.13"
 Functors = "0.1, 0.2, 0.3, 0.4"
 Parsers = "2"
-ProgressMeter = "1.2"
+ProgressMeter = "1.2, 1.3, 1.4, 1.5, 1.6, 1.7"
```
jeremiahpslewis, line 39:
```suggestion
ProgressMeter = "^1.2"
```

src/ReinforcementLearningCore/src/core/hooks.jl
jeremiahpslewis, line 120:
```suggestion
(h::TotalRewardPerEpisode)(::PostActStage, agent, env) =
```

HenriDeh:
Very cool showcase of speeding up with simple changes. Sorry I could not review the second time.

PR #830, opened 2023-03-09
Refactor and add tests to TotalBatchRewardPerEpisode

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?

### DIFF
--- a/src/ReinforcementLearningCore/src/core/hooks.jl
+++ b/src/ReinforcementLearningCore/src/core/hooks.jl
@@ -113,33 +113,32 @@ Base.getindex(h::RewardsPerEpisode) = h.rewards
 Store the total reward of each episode in the field of `rewards`. If
 `is_display_on_exit` is set to `true`, a unicode plot will be shown at the [`PostExperimentStage`](@ref).
 """
-struct TotalRewardPerEpisode{T, F} <: AbstractHook where {T<:Union{Bool,Nothing}, F<:Number}
+mutable struct TotalRewardPerEpisode{T,F} <: AbstractHook where {T<:Union{Bool,Nothing},F<:Number}
     rewards::Vector{F}
-    reward::Ref{F}
+    reward::F
     is_display_on_exit::Bool
 
-    function TotalRewardPerEpisode{F}(; is_display_on_exit::Bool = true) where {F<:Number}
+    function TotalRewardPerEpisode{F}(; is_display_on_exit::Bool=true) where {F<:Number}
         struct_type = is_display_on_exit ? Bool : Nothing
-        new{struct_type, F}([], Ref(convert(F, 0.0)), is_display_on_exit)
+        new{struct_type,F}([], 0.0, is_display_on_exit)
     end
 
-    function TotalRewardPerEpisode(; is_display_on_exit::Bool = true)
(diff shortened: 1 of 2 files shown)


src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -104,28 +104,28 @@ Base.getindex(h::RewardsPerEpisode) = h.rewards
 Store the total reward of each episode in the field of `rewards`. If
 `is_display_on_exit` is set to `true`, a unicode plot will be shown at the [`PostExperimentStage`](@ref).
 """
-mutable struct TotalRewardPerEpisode{T} <: AbstractHook where {T<:Union{Bool,Nothing}}
+struct TotalRewardPerEpisode{T} <: AbstractHook where {T<:Union{Bool,Nothing}}
     rewards::Vector{Float64}
-    reward::Float64
+    reward::Vector{Float64}
```
HenriDeh, line 109: I suppose the goal here is to let this be mutated. It is a better idea to use `reward::Ref{Float64}` then, because a vector allocates memory for an array, even when empty. Plus, it is a bit confusing to the reader why this is a one-element vector. It works with empty indexing:
image

jeremiahpslewis: Nice! I didn't know about this, thanks!

jeremiahpslewis: After looking into this a bit more, the most reasonable approach is leaving it the way it was, mutable keyword yields the desired mutability and doesn't cause performance issues.

src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -108,41 +108,44 @@ Base.getindex(h::RewardsPerEpisode) = h.rewards
...
     env,
 )
-    push!(hook.rewards, hook.reward[1])
-    hook.reward[1] = 0
+    push!(hook.rewards, hook.reward)
+    hook.reward = 0
 end
 
-function (hook::TotalRewardPerEpisode{Bool})(
+function (hook::TotalRewardPerEpisode{Bool, F})(
```
HenriDeh, line 144: Now that I think of it, instead of using `Bool` vs `Nothing`, we could use `Val{true}` and `Val{false}`

HenriDeh (dismissed):

jeremiahpslewis:
@HenriDeh Ok, one more trick: `Base.show` overloading so the hook itself prints the plot in the REPL (if the plot mode is activated). Thoughts?

src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -138,27 +140,33 @@ function (hook::TotalRewardPerEpisode)(
...
+        println(io, typeof(hook))
+    end
+end
+
+function (hook::TotalRewardPerEpisode{true, F})(
+    ::PostExperimentStage,
+    agent,
+    env,
+) where {F<:Number}
+    display(hook)
```
HenriDeh, line 161: I like the use of display here because some day we could let the plotting backend be user-configurable to instead use Plots or Makie.

HenriDeh (approved):

HenriDeh:
I made the merge

jeremiahpslewis:
Nice, thanks!

PR #834, opened 2023-03-11
MultiAgent Baseline Policies: Independent Learning / MADDPG

Mytolo:
As stated in the REAME.md of the new subproject of RL.jl, the MultiAgentManager was already present in the module. I adapted it to the current library run and call structure.
Independent Learning is a good baseline in order to compare algorithms against each other, and therefore should be present in a MultiAgent Part.
The subproject is sound since a lot of parts in the single agent case would be code duplicate for multi agent systems (MAS). Especially, this PR show that the single agent case is useful for MAS.
A great thanks to @peterchen96 for added some baseline algorithms in the first place.
Another pull request will be pushed in the next days, including Multi Agent Deep Dedtermnistic Policy Gradients (MADDPG), which was present in the version 0.10.0 of ReinforcementLerning.jl as well.


jeremiahpslewis:
@Mytolo Hey! This is great, I was hoping we could get multiagent back up and running. Can you add tests for the functions and structs you introduce in this PR?

src/MultiAgentReinforcementLearning/Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "MultiAgentReinforcementLearning"
+uuid = "c65af49a-feda-45df-9d14-b4e40efca6f2"
+authors = ["Panajiotis Keßler <panajiotis@christoforidis.net>"]
+version = "0.1.0"
+
+[deps]
+BSON = "fbb218c0-5317-5bc6-957e-2ee96dd4b1f0"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+ReinforcementLearningBase = "e575027e-6cd6-5018-9292-cdc6200d2b44"
```
jeremiahpslewis, line 10: Can you add Compat bounds as appropriate?

src/MultiAgentReinforcementLearning/README.md
```diff
@@ -0,0 +1,17 @@
...
+This package provides multi agent reinforcement learning baselines as well as some further techniques.
+
+Basically, the following parts are included in this module
+
+- `MultiAgentManager`
+- `MADDPG`
+- `DIAL`
+- `CoordinationGraphs`
+- `MessageContainer` for `ReinforcementLearningTrajectory` Module
+- `wholeEpisodeSampler` for `ReinforcementLearningTrajectory` Module
```
jeremiahpslewis, line 12: Should the capitalization here be consistent? E.g. ‘W’?

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,53 @@
+export MultiAgentManager
+
+struct MultiAgentManager <: AbstractPolicy
+    agents::Dict{Any,Any}
```
jeremiahpslewis, line 4: I never understood why in the original version, dict values can be type any…shouldn’t they be <: AbstractPolicy?

jeremiahpslewis (changes requested):
Added a few questions / comments.

src/MultiAgentReinforcementLearning/README.md
Mytolo: DIAL stands for Differentiable Inter Agent Learning as proposed by [
Learning to Communicate with Deep Multi-Agent Reinforcement Learning (Jakob N. Foerster, ...)](https://arxiv.org/abs/1605.06676). Hence, all the great letters should be fine.
For `WholeEpisodeSampler` I will change that in the `README.md`.

Mytolo:
> @Mytolo Hey! This is great, I was hoping we could get multiagent back up and running. Can you add tests for the functions and structs you introduce in this PR?

@jeremiahpslewis 
What exact;y do you mean? Interface tests as documentation of their use? Like Bound-tests and exception tests? Or do you also mean actual algorithm performance tests like MultiAgent with 10 Episodes should reach at least performance x? 
Would be great if you could provide some example. Thanks

jeremiahpslewis:
Something like these? https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/test/core/hooks.jl

https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/test/core/stop_conditions.jl

Right now we have the `RLExperiments.jl` which sort of function like smoke tests, but are very nonspecific. It would be good to know with a high degree of confidence that a given package 'works' & respects the expected / standard interface before we get to `RLExperiments.jl`, should also make issues quicker to diagnose.

Mytolo:
> Something like these? Not sure if they fit here; in a perfect world we'd already have tests for `Agent` and you could just copy them; I guess this time we'll do it the other way around, port your tests to `Agent` when ready ;). https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/test/core/hooks.jl
> 
> https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/test/core/stop_conditions.jl
> 
> Right now we have the `RLExperiments.jl` which sort of function like smoke tests, but are very nonspecific. It would be good to know with a high degree of confidence that a given package 'works' & respects the expected / standard interface before we get to `RLExperiments.jl`, should also make issues quicker to diagnose.

I also added an example for independent learning in `RLExperiments.jl` in this pull request.

src/MultiAgentReinforcementLearning/Project.toml
Mytolo: I am working on the newest versions of the Libraries. Shall I add them nonetheless?

Mytolo:
I will add both MultiAgentManager and MADDPGManager to the PR and appropriate tests.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: Changes in commit [606c3ba](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/834/commits/606c3baf4a539ec99ae2e6dec3dbd1ccb32af584)

src/MultiAgentReinforcementLearning/Project.toml
jeremiahpslewis: Yep! No need to pin to point versions, but at least to the minor release versions / ranges which are acceptable

src/MultiAgentReinforcementLearning/README.md
Mytolo: Changed in current commit of PR.

src/MultiAgentReinforcementLearning/Project.toml
Mytolo: I did the compat entries

Mytolo:
I need to adjust the test to the updates in the `main` Branch. Till now, they will not work with the agent refactor. (I was not aware of that change going to happen)
Are the tests in its current form suitable, @jeremiahpslewis ?

src/MultiAgentReinforcementLearning/src/policies/maddpg.jl
```diff
@@ -0,0 +1,157 @@
...
+- `update_freq::Int`
+- `update_step::Int`, count the step.
+- `rng::AbstractRNG`.
+"""
+mutable struct MADDPGManager{T <: AbstractPolicy} <: AbstractPolicy
+    agents::Dict{<:Any, T}
+    batch_size::Int
+    update_freq::Int
+    update_step::Int
+    rng::AbstractRNG
```
jeremiahpslewis, line 26: Same thing here, not sure what you see if you run `JET.jl` in terms of type stability, but it might make sense to use a param in `MADDPGManager` for the rng type.

src/MultiAgentReinforcementLearning/src/policies/maddpg.jl
```diff
@@ -0,0 +1,157 @@
...
+    # state
+    for i ∈ 1:length(first(batches)[2])
+      optimise!(π, Dict(player => batches[player][i] for (player, ) in π.agents))
+    end
+    
+end
+
+function RLBase.optimise!(π::MADDPGManager, player_batches)
+    # get s, a, s′ for critic
+    s = vcat((player_batches[player][:state] for (player, _) in π.agents)...)
```
jeremiahpslewis, line 94: Do you have a sense of what this does in terms of creating allocations? Is performance ok? (Naturally, we can merge this and then look at performance in a second step, but it's always good to get the API setup so that performance improvements are easier to implement.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,56 @@
+export MultiAgentManager
+
+Base.@kwdef mutable struct MultiAgentManager{T <: AbstractPolicy} <: AbstractPolicy
```
jeremiahpslewis, line 3: What about
```suggestion
Base.@kwdef mutable struct MultiAgentManager{T <: AbstractPolicy, S} <: AbstractPolicy
```

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,56 @@
...
+    agents::Dict{S,T} where {S<: Union{String, Symbol, Integer}}
```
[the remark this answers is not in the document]
jeremiahpslewis: and then

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,56 @@
...
+    cur_player::Union{String, Symbol, Integer} = first(keys(agents))
```
jeremiahpslewis, line 5: and
```suggestion
    cur_player::S = first(keys(agents))
```

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,56 @@
...
+Base.getindex(A::MultiAgentManager, x) = getindex(A.agents, x)
+
+"""
+    MultiAgentManager(player => policy...)
+
+This is the simplest form of multiagent system. At each step they observe the
+environment from their own perspective and get updated independently.
+"""
+MultiAgentManager(policies...) =
+    MultiAgentManager(Dict{Any,Any}(nameof(p) => p for p in policies))
```
jeremiahpslewis, line 17: Can we avoid allocating a `Dict{Any,Any}`? Can we get away with a `NamedTuple`?

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
```diff
@@ -0,0 +1,56 @@
...
+
+RLBase.prob(A::MultiAgentManager, env::AbstractEnv, args...) = prob(A[A.cur_player].policy, env, args...)
+
+(A::MultiAgentManager)(env::AbstractEnv) = A(env, DynamicStyle(env))
+
+function (A::MultiAgentManager)(env::AbstractEnv, ::Sequential)
+    while current_player(env) == chance_player(env)
+        env |> legal_action_space |> rand |> env
+    end
+    return A[A.cur_player](env)
```
jeremiahpslewis, line 27: It seems weird that in the `while` loop we fetch `current_player(env)`, but in the return statement we use `A.cur_player`, is there any reason for this inconsistency?

jeremiahpslewis (changes requested):
Sorry about the overlapping refactors, I was trying to chase down a few performance issues.

1. The new tests look great, thank you!!
2. Just a few comments about different parts of the `Agent` code; no need to worry about the `vcat` allocations if they aren't easily replaced, but I think the type comments might be worth getting sorted out (I may be missing something here)
3. Can you please rebase on the current main branch? There are a lot of commits from `main` which are showing up in the PR and make it hard to get through.

👏  for getting multi agent back in RL.jl!!

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: Actually, this part stayed as it was done by @peterchen96. In my opinion, we do not need it anyway. `NamedPolicy` is no part of the module as well.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo:
```suggestion
    A.cur_player = current_player(env)
    return A[A.cur_player](env)
```

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: @jeremiahpslewis do you think that it this constructor should be present for ease of use?

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: Nice suggestion. I changed it. Will be present in next version.

src/MultiAgentReinforcementLearning/src/policies/maddpg.jl
```diff
@@ -0,0 +1,157 @@
...
+  trajectory to update its own network. **Note that** here the policy of the `Agent` should be `DDPGPolicy` wrapped by `NamedPolicy`, see the relative 
+  experiment([`MADDPG_KuhnPoker`](https://juliareinforcementlearning.org/docs/experiments/experiments/Policy%20Gradient/JuliaRL_MADDPG_KuhnPoker/#JuliaRL\\_MADDPG\\_KuhnPoker) or [`MADDPG_SpeakerListener`](https://juliareinforcementlearning.org/docs/experiments/experiments/Policy%20Gradient/JuliaRL_MADDPG_SpeakerListener/#JuliaRL\\_MADDPG\\_SpeakerListener)) for references.
+- `traces`, set to `SARTS` if you are apply to an environment of `MINIMAL_ACTION_SET`, or `SLARTSL` if you are to apply to an environment of `FULL_ACTION_SET`.
+- `batch_size::Int`
+- `update_freq::Int`
+- `update_step::Int`, count the step.
+- `rng::AbstractRNG`.
+"""
+mutable struct MADDPGManager{T <: AbstractPolicy} <: AbstractPolicy
+    agents::Dict{<:Any, T}
```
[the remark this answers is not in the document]
Mytolo: Basically, it is the same as for `multi_agent.jl`. In some cases, it might be useful. (But I am not sure about that)
I will add it since it would be no limitation factor to performance.

src/MultiAgentReinforcementLearning/src/policies/maddpg.jl
Mytolo: Actually, I have not thought about allocation here. To avoid this, it would be necessary that every agent only updates its corresponding parts of the “whole” state. This makes the code a bit scrambled in order of readability. For sure, this is a fact one should have a look regarding performance.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
jeremiahpslewis: I'm now really confused...what is `current_player(env)` tracking vs `A.cur_player` from the manager? Seems like unnecessary duplication? Or a naming issue?

src/MultiAgentReinforcementLearning/src/policies/maddpg.jl
Mytolo: Actually, the range parameter is not needed here, since the samplers of the trajectories are doing the job. Previously, every policy needed to take care of the sampling on its own. This is not necessary anymore.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: I did it that way :+1:

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo:
```julia
Base.@kwdef mutable struct MultiAgentManager{T <: AbstractPolicy, S} <: AbstractPolicy
    agents::Dict{S,T}
    cur_player::S = first(keys(agents))
end
``` 

Base.@kwdef automatically defines the constructor. Hence, this part is not necessary in my opinion.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
Mytolo: In order to clarify it: `current_player(env)` is the agent which is going to take the next action. That being said, it means that after passing the action to the environment, it changes. However, `MultiAgentManager ` should update the elements of an agent's trajectory in the same iteration it took the action and passed it to the environment. Otherwise, the collected reward is a time shifted version and for the independent method this would mean that false reward signals would be present. 
The agent would not be able to learn an optimal policy (if that is possible for independent methods in general). 
Does that clear your confusion, or do I miss a point here?

Mytolo:
@jeremiahpslewis just a note: Are the RLExperiments tests adjusted to the new `Agent` Interface? It looks like they are all failing due to the unknown keyword arguments policy, trajectory.

src/MultiAgentReinforcementLearning/src/policies/multi_agent.jl
jeremiahpslewis: Thanks for your clarification, it really helped! What I wasn't really paying attention to is the degree to which the new `multiagent` setup differs from the old version, e.g. [here](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/v0.10.0/src/ReinforcementLearningCore/src/policies/agents/multi_agent.jl). I (personally) don't love that we now carry around the current player in two different places (e.g. `env` and `MultiAgent`) and that `chance_player` is being used wherever the `Sequential` trait exists...AFAIK, `chance_player` should only need to be implemented for non-`Deterministic` environments, but I don't see where this is supported in the new code.

jeremiahpslewis:
> @jeremiahpslewis just a note: Are the RLExperiments tests adjusted to the new `Agent` Interface? It looks like they are all failing due to the unknown keyword arguments policy, trajectory.

Yes, sadly somehow that wasn't caught by the tests pre-merge, will be fixed when https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/849 is merged later today / tomorrow.

HenriDeh:
#849 is merged you shouldn't have the issue anymore. Sorry I have the bad habit of doing stuff in big PRs when I could've fixed this in a quick separate one.

jeremiahpslewis (changes requested):
@Mytolo Hey! Sorry that this PR is taking so long, but it turns out to be a bit more complicated to resolve. :) I put together a very basic multiagent test based on some of the earlier quick start documentation and would like to know what you think...

1. Do we need a test like this?
2. Why does it fail?
3. Can we drop the current_player from `MultiAgent` object? It seems like we keep putting more state in `Agent` objects which might better be kept / duplicates elements already present in the `Environment`, and the more attributes / objects we introduce, the harder it is going to be to keep track of them.

jeremiahpslewis:
@Mytolo Would something like this (https://github.com/Mytolo/ReinforcementLearning.jl/pull/1) be viable?

Mytolo:
Sorry for answering late. 
1. Regarding the `TicTacToe` test. What exactly should be tested here? Or is it to have another environment different from `pettingzoo` environments? In my opinion, the provided tests are sufficient.
2. To be honest, I do not know which part of my work triggers that, since I did not change anything in this PR concerning RLCore. 
3. On my side, it does not fail as it seems at the moment. However, it looks like a memory issue with the GitHub pipeline. Or does it fail on your side as well?

Mytolo:
> @Mytolo Would something like this ([Mytolo#1](https://github.com/Mytolo/ReinforcementLearning.jl/pull/1)) be viable?

This is not working as it should, as I said. The rewards and states of the players get scrambled. Hence, the independent single agent RL method would get false signals. (Learning anything but nothing useful)

Mytolo:
@jeremiahpslewis would it be viable to extend the curent_player(env) with a keyword argument like `takes_action=true`? The default would be to get the current player taking the next action.  `false` means that it is the player in the current step nonetheless of the action being passed to the environment or not.

Mytolo:
Since this PR is quite old, I am going to rebase everything. As far as I have noticed, adding a MultiAgent Subfolder to the project is not wanted? How should i provide multi agent policies? 
In a subfolder of ReinforcementLearningZoo? (For example multiagent)

HenriDeh:
> Since this PR is quite old, I am going to rebase everything. As far as I have noticed, adding a MultiAgent Subfolder to the project is not wanted? How should i provide multi agent policies? In a subfolder of ReinforcementLearningZoo? (For example multiagent)

As I understand it, things that are algorithm specific should go into the Zoo package, while modules that can be used to construct algorithms and policies should go in the Core package.

jeremiahpslewis:
@Mytolo I'm going to close this PR. Want to pick one of the components of this PR to prioritize and put it in a follow up PR (once #880 is merged)?

PR #848, opened 2023-03-15
Improve 3D mvnormlogpdf and namespace

HenriDeh:
PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningCore/src/utils/distributions.jl
+++ b/src/ReinforcementLearningCore/src/utils/distributions.jl
@@ -1,6 +1,5 @@
 export normlogpdf, mvnormlogpdf, diagnormlogpdf, mvnormkldivergence, diagnormkldivergence, normkldivergence
 
-using Flux: unsqueeze, stack
 using LinearAlgebra
 
 # watch https://github.com/JuliaStats/Distributions.jl/issues/1183
@@ -63,8 +62,9 @@ action_samples x batch_size).  Return a 3D matrix of size (1 x action_samples x
 batch_size). 
 """
 function mvnormlogpdf(μ::A, LorU::A, x::A; ϵ=1.0f-8) where {A<:AbstractArray}
-    logp = [mvnormlogpdf(μ[:, :, k], LorU[:, :, k], x[:, :, k]) for k in 1:size(x, 3)]
-    return unsqueeze(stack(logp; dims=2), dims=1) #returns a 3D vector 
+    l(k) = mvnormlogpdf(μ[:, :, k], LorU[:, :, k], x[:, :, k])
+    logp = reduce(hcat, [l(k) for k in axes(x, 3)])
+    return reshape(logp, 1, size(logp)...)
 end
 
 #Used for mvnormlogpdf
--- a/src/ReinforcementLearningCore/src/utils/distributions.jl
+++ b/src/ReinforcementLearningCore/src/utils/distributions.jl
@@ -62,7 +62,7 @@ batch_size).
 """
 function mvnormlogpdf(μ::A, LorU::A, x::A; ϵ=1.0f-8) where {A<:AbstractArray}
     l(k) = mvnormlogpdf(μ[:, :, k], LorU[:, :, k], x[:, :, k])
-    logp = reduce(hcat, [l(k) for k in axes(x, 3)])
+    logp = reduce(hcat, (l(k) for k in axes(x, 3)))
     return reshape(logp, 1, size(logp)...)
 end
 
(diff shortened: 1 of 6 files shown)


HenriDeh:
@jeremiahpslewis What are the tests you felt were missing for mvnormlogpdf ?

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/ppo.jl
```diff
@@ -1,4 +1,5 @@
 export PPOPolicy, PPOTrajectory, MaskedPPOTrajectory
+import ReinforcementLearningCore.normlogpdf
```
jeremiahpslewis, line 2: Should we make this consistent with the previous example and do:
```suggestion
import ReinforcementLearningCore: normlogpdf
```

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/vmpo.jl
```diff
@@ -2,6 +2,7 @@ export VMPOPolicy, VMPOTrajectory
 
 using LinearAlgebra: ⋅  # dot product
 using ChainRulesCore: ignore_derivatives
+import ReinforcementLearningCore.normlogpdf
```
jeremiahpslewis, line 5: Should we make this consistent with the first example and do:
```suggestion
import ReinforcementLearningCore: normlogpdf
```

jeremiahpslewis (changes requested):
Just have a few ideas for minor tweaks, let me know what you think!

src/ReinforcementLearningCore/src/utils/distributions.jl
```diff
@@ -63,13 +61,15 @@ action_samples x batch_size).  Return a 3D matrix of size (1 x action_samples x
 batch_size). 
 """
 function mvnormlogpdf(μ::A, LorU::A, x::A; ϵ=1.0f-8) where {A<:AbstractArray}
-    logp = [mvnormlogpdf(μ[:, :, k], LorU[:, :, k], x[:, :, k]) for k in 1:size(x, 3)]
-    return unsqueeze(stack(logp; dims=2), dims=1) #returns a 3D vector 
+    l(k) = mvnormlogpdf(μ[:, :, k], LorU[:, :, k], x[:, :, k])
+    logp = reduce(hcat, [l(k) for k in axes(x, 3)])
```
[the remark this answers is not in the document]
HenriDeh: Yes, this is really tricky to optimize, I've tried quite a lot to minimize allocations. The thing is, this must be differentiable (no array mutation as long as we work with Zygote), and GPU compatible (no array indexing). I was told that reduce hcat on an array comprehension is well optimized by the compiler. 

Now that I write this, I see that there is array indexing here. I'll run some tests locally.

jeremiahpslewis:
Just going through the different functions, a question came up...this version is faster on a CPU, would it work on a GPU?
```julia
function logdetLorU(LorU::CuArray)
    return 2*log(prod(diag(LorU)))
end
```

jeremiahpslewis:
And what about pulling out the `2` from `logdetLorU` and using `const log4π` instead? Seems to speed things up by nearly 10%.

src/ReinforcementLearningCore/src/utils/distributions.jl
jeremiahpslewis, line 65: Want to use a tuple here instead of a vect?

```suggestion
    logp = reduce(hcat, (l(k) for k in axes(x, 3)))
```

HenriDeh:
> And what about pulling out the `2` from `logdetLorU` and using `const log4π` instead? Seems to speed things up by ~5%.

Do you mean replacing the function by "half_logdetLorU" and multiplying its output by two when called ?

HenriDeh:
> And what about pulling out the `2` from `logdetLorU` and using `const log4π` instead? Seems to speed things up by ~5%.

Do you mean in replacing the log2pi in mvnormlogpdf? Because `a*log(2pi) + logdet != a*log(4pi) + logdet/2` so I'm not sure what you are suggesting.

jeremiahpslewis:
> > And what about pulling out the `2` from `logdetLorU` and using `const log4π` instead? Seems to speed things up by ~5%.
> 
> Do you mean in replacing the log2pi in mvnormlogpdf? Because `a*log(2pi) + logdet != a*log(4pi) + logdet/2` so I'm not sure what you are suggesting.

Yup, let's ignore that idea. 😆

HenriDeh:
> Just going through the different functions, a question came up...this version is faster on a CPU, would it work on a GPU?
> 
> ```julia
> function logdetLorU(LorU::CuArray)
>     return 2*log(prod(diag(LorU)))
> end
> ```

It does not. 
image
Anyways, the CPU fallback below is much faster as it uses the built-in logdet function of LinearAlgebra.

PR #850, opened 2023-03-16
Agent Refactor

jeremiahpslewis:
Latest timings of `core.jl` `run` function tests:

New: 7.864 μs (158 allocations: 6.95 KiB)
Old: 161.917 μs (260 allocations: 12.02 KiB)

Requires: https://github.com/JuliaReinforcementLearning/ReinforcementLearningTrajectories.jl/pull/36

The goal is to eliminate allocations and increase the degree of type stability (not entirely type stable, but the issues are in `RLTrajectories.jl`, so not relevant for this PR.

This PR structures the `Agent` `cache` object as having `SART` fields, which makes the `cache` type stable. In the previous version, the `cache` changed type signature as different elements were included / dropped from the `NamedTuple`.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis:
@HenriDeh I don't know if the changes I made here really make sense, but if they do make sense, we get 0 allocations and a 10-20x speedup for key operations.

HenriDeh:
Yeah I was think that Agent was highly type unstable with the way the cache is implemented. Give me a moment to review.

src/ReinforcementLearningCore/src/policies/agent.jl
```diff
@@ -13,21 +54,27 @@ is a Callable and its call method accepts varargs and keyword arguments to be
...
-        agent = new{P,T}(policy, trajectory, cache)
         if TrajectoryStyle(trajectory) === AsyncTrajectoryStyle()
             bind(trajectory, @spawn(optimise!(policy, trajectory)))
         end
         agent
     end
 end
 
-Agent(; policy, trajectory, cache=NamedTuple()) = Agent(policy, trajectory, cache)
+Agent(; policy, trajectory, env=missing) = Agent(policy, trajectory, cache)
```
HenriDeh, line 77: Isn't cache not defined here ?

src/ReinforcementLearningCore/src/policies/agent.jl
```diff
@@ -13,21 +54,27 @@ is a Callable and its call method accepts varargs and keyword arguments to be
...
     trajectory::T
-    cache::NamedTuple # trajectory do not support partial inserting
+    cache::C # trajectory does not support partial inserting
+
+    function Agent(policy::P, trajectory::T; env=missing) where {P,T}
+        if !ismissing(env)
+            cache = AgentCache(policy, env)
+            agent = new{P,T, typeof(cache)}(policy, trajectory, cache)
+        else
+            agent = new{P,T, typeof(AgentCache())}(policy, trajectory, AgentCache())
```
HenriDeh, line 67: Another possibility is to use the trajectory to infer the type of the cache's state and action types. By default a trajectory will assume they are Integers. But if you have Vector states you can specify it when constructing the trajectory.

HenriDeh (commented):
I'll finish review tomorrow with a fresh mind

jeremiahpslewis:
@HenriDeh Looking forward to your review tomorrow. I've pushed a second approach to the refactor which seems to resolve things a bit better. :)

src/ReinforcementLearningCore/src/policies/agent.jl
```diff
@@ -4,6 +4,37 @@ using Base.Threads: @spawn
...
+    action::Union{A, Missing}
+    reward::Union{R, Missing}
+    terminal::Union{Bool, Missing}
+
+    function SART()
+        new{Any, Any, Any}(missing, missing, missing, missing)
+    end
+end
+
+sart_to_tuple(sart::SART{S,A,R}) where {S,A,R} = @NamedTuple{state::Union{S,Missing}, action::Union{A,Missing}, reward::Union{R,Missing}, terminal::Union{Bool,Missing}}((sart.state, sart.action, sart.reward, sart.terminal))
```
HenriDeh, line 18: Maybe we could overload Base.tuple instead ? To limit new names.

HenriDeh (commented):
This goes in the right direction. However I foresee one issue: Agent is now specialized fort SART algorithms and cannot be modulated to work with other types of traces, or even custom traces. For example, some algorithms need to store action masks or actoion logprobabilities. Ideally, cache would be a parametrically typed to accept SART, but also other cache types (maybe with a shared supertype). We do not need to define such other tyoes now, this can be done when needed.

jeremiahpslewis:
@HenriDeh after three versions, think this is close to done!

jeremiahpslewis:
Let me know. :)

src/ReinforcementLearningCore/src/policies/agent_srt_cache.jl
```diff
@@ -0,0 +1,38 @@
+mutable struct SRT{S,R,T}
```
HenriDeh, line 1: I can't help but wonder if using Union types is not detrimental to performance ? Sure it's better the the former NamedTuple of Anys. But I was wondering if there would not be a way to initialize this trace to undef with a boolean field that asserts whether the cache still has undef variables. Thus, we wouldn't need to check for isnothing at all (simply set the field to true at the PostActStage).

src/ReinforcementLearningCore/test/policies/agent.jl
```diff
@@ -0,0 +1,93 @@
...
+        push!(trajectory, srt_3, 2)
+        @test length(trajectory.container) == 1
+        @test trajectory.container[:action] == [1]
+        push!(trajectory, srt_3, 3)
+        @test trajectory.container[:action] == [1, 2]
+        @test trajectory.container[:state] == [1, 1]
+    end
+
+    @testset "Agent Tests" begin
+        a_1 = Agent(
```
HenriDeh, line 40: When constructing an agent without explicitly supplying the cache struct, it will make one that has `Union{Any,Nothing} fields. This remains a perfomance problem I believe. Could we not insert in the `Agent(policy, trajectory)` constructor a routine that infers the type of each Trace based on the trajectory, and construct a typed cache. This information was already supplied by the user when constructing the trajectory. So it exists somewhere.

src/ReinforcementLearningCore/src/policies/agent_srt_cache.jl
```diff
@@ -0,0 +1,38 @@
...
+function update!(cache::SRT{S,R,T}, state::S) where {S,R,T}
+    cache.state = state
+end
+
+function update!(cache::SRT{S,R,T}, reward::R, terminal::T) where {S,R,T}
+    cache.reward = reward
+    cache.terminal = terminal
+end
+
+function RLBase.reset!(cache::SRT)
```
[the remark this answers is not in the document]
HenriDeh: Well, I guess it is useful to reset at the end of the experiment if one wants to reuse the agent to resume training.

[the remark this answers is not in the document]
jeremiahpslewis: For benchmarking runs it can be helpful; this only runs once per experiment, so no meaningful cost.

src/ReinforcementLearningCore/src/policies/agent_srt_cache.jl
jeremiahpslewis: This isn't a union type, it's a type parameter for each of `S`, `R` and `T`, but each is type-pure.

src/ReinforcementLearningCore/test/policies/agent.jl
jeremiahpslewis: So far, I haven't seen a meaningful performance hit for not specifying the cache, so I left it the way it is.

src/ReinforcementLearningCore/src/policies/agent.jl
```diff
@@ -4,6 +4,37 @@ using Base.Threads: @spawn
...
+
+function RLBase.reset!(sart::SART)
```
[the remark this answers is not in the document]
jeremiahpslewis: Correct, I agree. Reset should only happen at end of experiment.

src/ReinforcementLearningCore/test/policies/agent.jl
HenriDeh: If you say so then ok. It certainly won't be worse than before.

HenriDeh (approved):

jeremiahpslewis:
@HenriDeh Thanks for the careful review!

Issue #843, opened 2023-03-14
test/runtests.jl empty (+ arch discussion)


filchristou:
Quite schockingly, I realised `test/runtests.jl` is empty. 
What's the current way to test this package ?

jeremiahpslewis:
Each submodule has tests, see [here](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/.github/workflows/ci.yml) for what gets tested by Github Actions.

jeremiahpslewis:
I don't think so, but running it off of a branch / PR on Github works.

HenriDeh:
You can activate a subrepo in the REPL then run the tests like this: `]activate ReinforcementLearningCore`, `]test ReinforcementLearningCore` (for the case of RLCore). Testing the entirety of the metapackage requires you to do this one by one. Maybe we could automate this. Careful that the structure of RL.jl will make it such that the released versions of the dependencies will be used and not your local one. So if you made changes to RLBase and want to test RLCore, do not forget to `]dev src/ReinforcementLearningBase`

filchristou:
> Maybe we could automate this. 

What if we create a separate environment in `ReinforcementLearning.jl/test/` that will have its own `Manifest.toml` `dev`ing all the local passages. I myself am a fan of this approach and it might fit well (some discussion here https://julialang.zulipchat.com/#narrow/stream/225583-appreciation/topic/Run.2Ejl)

jeremiahpslewis:
Want to put together a PR with some instructions on how to use it? Happy to try it out!

filchristou:
I am rethinking a bit what are the requirements and scenarios to have such a local deving/testing machinery.

# Summary
My understanding is that the general ReinforcementLearning (RL) module/package is just the umbrella handler 
of the (most recent) eligible version combinations of the submodules.
Basically developing the RL.jl means updating the Project.toml versioning.
This means a specific RL installed instance (`add RL`) most possibly uses submodules from different RL.jl versions.

# Scenarios
## Developing
When the user wants to make certain changes in the code base.

### 1. Developing a specific submodule (e.g. `RLBase`, `RLCore`, `RLEnvironments`, ...)
The user specifically wants to alter the functionality for a specific submodule.
#### Implementation and caveats
We must `dev` the specific submodule and **let all others to default**, i.e. the general registry.
`dev`ing a submodule will **freeze** the version dependency to what currently is located in the `Project.toml`.
[...]

PR #853, opened 2023-03-24
WIP: faciliate local testing

filchristou:
An early draft of what I meant in https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/issues/843#issuecomment-1483232391

At the moment doing  `$> julia --project ./test` and then `julia> include(./test/runtests.jl)` will succesfully test ReinforcementLearningCore and ReinforcementLearningBase.

I had some problems with ReinforcementLearningEnvironments, but I think with time I could (probably) get it to work.

I think this approach might be interesting. I will be investigating it further, but firstly I would like to share the idea in case it's a no-go for you, so that nobody wastes time.
I had some problems with `dev`ing DistributedReinforcementLearning and ReinforcementLearningExperiments.. still not sure why.
It's rather scary seeing all these dependencies together.

So please, although it's non breaking and harmless, don't merge still!


jeremiahpslewis:
@filchristou Don't include DistributedRL for the moment? Not sure that it has been migrated to the latest syntax, etc. I would only include the sub-packages used in demoed, here (https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/devmode.jl)

jeremiahpslewis:
Closing this for now, feel free to reopen.

filchristou:
Hi. Sorry for the delay. I believe I will find the time before juliacon to seriously engage with this.

filchristou:
Some related questions [here](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/commit/395f23fa94e1f5d5dba342c19feb7b4f892030d1#r119239391)

PR #854, opened 2023-03-28
Include codecov in CI

HenriDeh:


HenriDeh:
@jeremiahpslewis I took a look at codecov to see what it entails to include this workflow. This promises to be a pain due to the nested structure of this package. We would have to see how to use the julia-coverage action while keeping the dependencies dev'ed.

jeremiahpslewis (commented):
Ready to go, drop second Julia version check

jeremiahpslewis:
@HenriDeh This seems to work now and I've had the codecov settings updated so that we get the PR comments (thanks @findmyway!). I've also dropped version=1 testing in the CI script as I think we really only want to focus on supporting the latest full Julia release (until we get things stabilized and then we can think about broadening this).

PR #856, opened 2023-03-28
DoEvery should not trigger on t = 1

HenriDeh:
Currently, DoEvery hooks trigger at the first step (because `x % 1 == 0`). But unless we want them to trigger every step/episode, we do not want them to trigger at t = 1.

PR Checklist

- [x] Update NEWS.md?
- [x] PR Review?


HenriDeh:
Last commit should be "Also test that it works as expected when t =1"

jeremiahpslewis (approved):

HenriDeh:
You're right, I was tired yesterday. I don't know why it does not work as intendented in my usecase. I'll investigate more.

jeremiahpslewis:
I'll merge this though, it's a great tweak!

PR #858, opened 2023-03-30
Tweak CI to check on dep changes

HenriDeh:
Currently CI is setup to only check subpackages in which changes occurred. But a change in RLCore might break algorithms in Zoo, for example. Currently this would not be caught by CI. This tweak should make sure to trigger tests of subpackages if a dependency changed. 

I will make an impactless change to RLCore n the second commit to see if correctly triggers Envs and Zoo tests, it shouldn't trigger Base.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,17 +159,23 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
+      - uses: marceloprado/has-changed-path@v1
+            id: documentation-changed
+            with:
+              paths: docs
       - uses: actions/checkout@v3
       - name: Setup python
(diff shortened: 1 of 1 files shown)


HenriDeh:
Documentation CI is lengthy too. We could add that if no changes were done to the docs, the tests shouldn't run?

jeremiahpslewis:
> Documentation CI is lengthy too. We could add that if no changes were done to the docs, the tests shouldn't run?

Sure!

jeremiahpslewis (dismissed):

jeremiahpslewis (approved):

HenriDeh:
Sorry for the spam but for some reason the action that checks whether the docs folder changed disables the whole CI. I don't understand.

jeremiahpslewis (approved):

PR #861, opened 2023-04-15
MultiAgent Proposal

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis:
@HenriDeh @Mytolo I'll put more work into cleaning this up starting end of April, but I wanted to get early feedback about the general direction and api.

Cheers,
Jeremiah

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+
+using ReinforcementLearningBase
+using ReinforcementLearningCore
+
+import ReinforcementLearningCore: RLCore
+import Base.getindex
+import Base.iterate
+
+struct MultiAgentPolicy <: AbstractPolicy
+    agents::NamedTuple
```
HenriDeh, line 14: This is going to be dynamically dispatched as `NamedTuple` is not a concrete type.
```suggestion
struct MultiAgentPolicy{NT<: NamedTuple} <: AbstractPolicy
    agents::NT
```

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+
+    function MultiAgentPolicy(agents::NamedTuple)
+        new(agents)
+    end
+end
+
+struct MultiAgentHook <: AbstractHook
+    hooks::NamedTuple
```
HenriDeh, line 22: Ibid

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+    function MultiAgentHook(hooks::NamedTuple)
+        new(hooks)
+    end
+end
+
+
+(p::MultiAgentPolicy)(env::AbstractEnv) = nothing # Default does nothing, but might be useful for some environments to clean up / pass final state to agents
+
+struct CurrentPlayerIterator
+    env::AbstractEnv
```
HenriDeh, line 33: Ibid
```suggestion
struct CurrentPlayerIterator{E<:AbstractEnv}
    env::E
```

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+end
+
+function current_player_iterator(env)
```
HenriDeh, line 36: I would avoid creating a second constructor unless there's a specific reason to do it.

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+    return CurrentPlayerIterator(env)
+end
+
+Base.iterate(current_player_iterator::CurrentPlayerIterator) =
+    (current_player(current_player_iterator.env), current_player_iterator.env)
+
+Base.iterate(current_player_iterator::CurrentPlayerIterator, env) =
+    (current_player(current_player_iterator.env), current_player_iterator.env)
```
HenriDeh, line 44: These seem to do the same, env is not needed. We can simplify to 
```suggestion

Base.iterate(current_player_iterator::CurrentPlayerIterator, env = current_player_iterator.env) =
    (current_player(current_player_iterator.env), env)
```
would allow a specific env argument, otherwise behaves as the two above functions.

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+Base.getindex(h::MultiAgentHook, s::Symbol) = h.hooks[s]
+
+Base.keys(p::MultiAgentPolicy) = keys(p.agents)
+Base.keys(p::MultiAgentHook) = keys(p.hooks)
+
+function RLCore._run(
+    multiagent_policy::MultiAgentPolicy,
+    env::AbstractEnv,
+    stop_condition,
+    hook::MultiAgentHook,
```
HenriDeh, line 59: Is it necessary to impose a `MultiAgentHook` when working with a MultiAgent ?

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+
+                policy(PostActStage(), env)
+                hook(PostActStage(), policy, env)
+
+                if stop_condition(policy, env)
+                    is_stop = true
+                    policy(PreActStage(), env)
+                    hook(PreActStage(), policy, env)
+                    multiagent_policy(env)  # let the policy see the last observation
+                    break
```
HenriDeh, line 162: This does not break the reset_condition while loop does it? Shouldn't it ?

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
```diff
@@ -0,0 +1,224 @@
...
+            multiagent_policy(PostEpisodeStage(), env)  # let the policy see the last observation
+            hook(PostEpisodeStage(), multiagent_policy, env)
+        end
+    end
+    multiagent_policy(PostExperimentStage(), env)
+    hook(PostExperimentStage(), multiagent_policy, env)
+    hook
+end
+
+function _run(
```
HenriDeh, line 177: What is the difference between this `_run` and the original one in RLBase? Does the RLBase one not already dispatch to this?

HenriDeh (commented):
I made a partial review of the new module, but I don't have the time for the rest today.

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
jeremiahpslewis, line 22:
```suggestion
    hooks::NamedTuple::NT
```

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
jeremiahpslewis: It does [here](https://docs.julialang.org/en/v1/manual/control-flow/#man-loops), but I know it's a bit intransparent. Copied from the previous `run` function.

src/MultiAgentReinforcementLearning/src/MultiAgentReinforcementLearning.jl
jeremiahpslewis: Great point!

src/ReinforcementLearningCore/src/policies/agent/multi_agent.jl
```diff
@@ -0,0 +1,184 @@
+export MultiAgentPolicy
+export MultiAgentHook
+
+using Random # for RandomPolicy
+
+import Base.getindex
+import Base.iterate
+
```
HenriDeh, line 8: This struct and those below in this file need at least a docstring to explain what they do and what arguments are expected by the constructors.

HenriDeh (changes requested):
Looks good. The only thing I think is missing is documentation. I think it is better to add it now rather than in an hypothetical future big doc PR. You already have examples you can link to. It would only require a small page that briefly explain what to do if you want to train on multiagents environments. That plus a docstring for the structs/functions that are user-level.

HenriDeh (dismissed):

jeremiahpslewis:
@HenriDeh Thank you for the great feedback! Can you re-approve?

PR #867, opened 2023-04-26
Multiplayer Fixes (Clean up errors)

jeremiahpslewis:
PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -62,12 +62,12 @@ Base.show(io::IO, m::MIME"text/plain", t::Experiment{S}) where {S} = show(io, m,
 Base.run(ex::Experiment) = run(ex.policy, ex.env, ex.stop_condition, ex.hook)
 
 function Base.run(
-    policy::P,
-    env::E,
+    policy::AbstractPolicy,
+    env::AbstractEnv,
     stop_condition=StopAfterEpisode(1),
     hook=EmptyHook(),
     reset_condition=ResetAtTerminal()
-) where {P<:AbstractPolicy, E<:AbstractEnv}
+)
     policy, env = check(policy, env)
     _run(policy, env, stop_condition, hook, reset_condition)
 end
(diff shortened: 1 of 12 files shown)


src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -62,21 +62,20 @@ Base.show(io::IO, m::MIME"text/plain", t::Experiment{S}) where {S} = show(io, m,
...
 function Base.run(
-    policy::AbstractPolicy,
-    env::AbstractEnv,
+    policy::P,
+    env::E,
     stop_condition=StopAfterEpisode(1),
     hook=EmptyHook(),
     reset_condition=ResetAtTerminal()
-)
+) where {P<:AbstractPolicy, E<:AbstractEnv}
```
HenriDeh, line 70: Why this? They should be strictly equivalent if I'm not mistaken. If there is no reason, I'd prefer we keep the old syntax as it is the canonical Julia syntax (we don't need E or P in the namespace).

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -62,21 +62,20 @@ Base.show(io::IO, m::MIME"text/plain", t::Experiment{S}) where {S} = show(io, m,
...
     policy, env = check(policy, env)
     _run(policy, env, stop_condition, hook, reset_condition)
 end
 
 "Inject some customized checkings here by overwriting this function"
 check(policy, env) = policy, env
 
-function _run(policy::AbstractPolicy, env::AbstractEnv, stop_condition, hook, reset_condition)
-
+function _run(policy::P, env::E, stop_condition, hook, reset_condition) where {P<:AbstractPolicy, E<:AbstractEnv}
```
HenriDeh, line 78: ibid

jeremiahpslewis:
@HenriDeh Thanks for the input, I guess I got the AbstractStruct-level specialization which you pointed out in the previous PR confused with function dispatch!

HenriDeh (approved):

PR #876, opened 2023-05-03
Add back ComposedStop

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningCore/src/core/stop_conditions.jl
+++ b/src/ReinforcementLearningCore/src/core/stop_conditions.jl
@@ -3,6 +3,28 @@ export StopAfterStep,
 
 import ProgressMeter
 
+
+#####
+# ComposedStopCondition
+#####
+
+"""
+    ComposedStopCondition(stop_conditions...; reducer = any)
+
+The result of `stop_conditions` is reduced by `reducer`.
+"""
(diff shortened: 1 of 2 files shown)


src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -1,8 +1,30 @@
...
+
+#####
+# ComposedStopCondition
+#####
+
+"""
+    ComposedStopCondition(stop_conditions...; reducer = any)
+
+The result of `stop_conditions` is reduced by `reducer`.
+"""
```
HenriDeh, line 15:
```suggestion
The result of `stop_conditions` is reduced by `reducer`. The default `reducer` is the `any` function, which means that the condition is true when any one of the `stop_conditions...` is true. Can be replaced by any function returning a boolean. For example `reducer = x->sum(x) >= 2` will require at least two of the conditions to be true.
"""
```
I added an extra example for the sake of clarity.

HenriDeh (commented):
Sorry I'm a little busy lately. I will be more available soon to review faster. 
I suggested a small change otherwise no comments.

HenriDeh (approved):

PR #879, opened 2023-05-04
Further refinements

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -69,7 +69,7 @@ function Base.run(
     env::AbstractEnv,
     stop_condition::AbstractStopCondition=StopAfterEpisode(1),
     hook::AbstractHook=EmptyHook(),
-    reset_condition=ResetAtTerminal()
+    reset_condition::AbstractResetCondition=ResetAtTerminal()
 )
     policy, env = check(policy, env)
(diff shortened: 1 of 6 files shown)


src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -34,10 +34,10 @@ function try_parse_kw(s)
...
 
 struct Experiment{S}
-    policy::Any
-    env::Any
-    stop_condition::Any
-    hook::Any
+    policy::AbstractPolicy
+    env::AbstractEnv
+    stop_condition::AbstractStopCondition
+    hook::AbstractHook
```
HenriDeh, line 40: While we're at it, lets do this the idiosyncractic way and not use abstract-typed fields (see perfomance tips of julia doc).
```
struct Experiment{P<: AbstractPolicy,E<: ect... ,S,H}
```
where {P,E,S,H} are the type fields. I think we should also add a reset_condition field. But you don't have to do it in this PR if it's too much.

jeremiahpslewis: Agree 100%, thanks for the nudge! Will add this to the PR.

HenriDeh:
Sure, it's not a big deal anyways as Experiment is typically not a struct that must be dispatched on frequently.

HenriDeh (approved):

PR #880, opened 2023-05-09
Use multiple dispatch / methods plan! and act!

jeremiahpslewis:
RLTrajectories has significant type instability 'challenges'. This PR works around them for a common use case, SART trajectories.

PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis:
@HenriDeh i know this doesn't pass the env tests yet, but before I migrate all of the env test cases to a new syntax, would you mind taking a look? If I'm not utterly mistaken, we get much better type stability (and performance speedups) via these changes as well as making it easier to search for specific methods.

src/ReinforcementLearningBase/src/base.jl
```diff
@@ -104,8 +104,8 @@ function test_interfaces!(env)
             A, A′ = legal_action_space(X), legal_action_space(Y)
```
HenriDeh, line 104: I know this is unrelated to your PR but shouldn't these tests be in the test directory? Can you perhaps copy paste them in a new file there, otherwise I suspect they'll stay here a while longer.

src/ReinforcementLearningBase/src/interface.jl
```diff
@@ -19,7 +19,7 @@ import Markdown
 #####
 
 """
-    (π::AbstractPolicy)(env) -> action
+    plan!(π::AbstractPolicy, env) -> action
 
 Policy is the most basic concept in reinforcement learning. Unlike the
 definition in some other packages, here a policy is defined as a functional
```
HenriDeh, line 25: This isn't true anymore the above change is made.

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -83,39 +83,41 @@ function _run(policy::AbstractPolicy,
...
 
-            if stop_condition(policy, env)
+            if stop(stop_condition, policy, env)
                 is_stop = true
-                policy(PreActStage(), env)
-                hook(PreActStage(), policy, env)
-                policy(env)  # let the policy see the last observation
+                update!(policy, PreActStage(), env)
+                update!(hook, PreActStage(), policy, env)
+                RLBase.plan!(policy, env)  # let the policy see the last observation
```
HenriDeh, line 110:
```suggestion
                update!(policy, PreActStage(), env) # let the policy see the last observation
                update!(hook, PreActStage(), policy, env)
                RLBase.plan!(policy, env)  # let the policy see the last action
```

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -83,39 +83,41 @@ function _run(policy::AbstractPolicy,
...
                 break
             end
         end # end of an episode
 
         if is_terminated(env)
-            policy(PostEpisodeStage(), env)  # let the policy see the last observation
-            hook(PostEpisodeStage(), policy, env)
+            update!(policy, PostEpisodeStage(), env)  # let the policy see the last observation
+            update!(hook, PostEpisodeStage(), policy, env)
```
HenriDeh, line 117: Shouldn't we add a `plan!` line here to see the last action as well ?

src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -23,8 +23,8 @@ struct ComposedStopCondition{S,T} <: AbstractStopCondition
     end
 end
 
-function (s::ComposedStopCondition)(args...)
-    s.reducer(sc(args...) for sc in s.stop_conditions)
+function stop(s::ComposedStopCondition, args...)
```
HenriDeh, line 26: Imho, the name `stop` does not convey what this function does. Can we perhaps rename this to `check_stop` or `check_stop_condition`?

src/ReinforcementLearningCore/src/policies/agent/multi_agent.jl
```diff
@@ -175,38 +175,54 @@ function Base.run(
...
-function (multiagent::MultiAgentPolicy)(::PostActStage, env::E) where {E<:AbstractEnv}
+function update!(multiagent::MultiAgentPolicy, ::PostActStage, env::E) where {E<:AbstractEnv}
     for player in players(env)
-        RLCore.update!(multiagent[player].cache, reward(env, player), is_terminated(env))
+        update!(multiagent[player].cache, reward(env, player), is_terminated(env))
     end
 end
 
-function (multiagent::MultiAgentPolicy)(::PostEpisodeStage, env::E) where {E<:AbstractEnv}
+function update!(multiagent::MultiAgentPolicy, ::PostExperimentStage, env::E) where {E<:AbstractEnv}
```
HenriDeh, line 196: `PostEpisodeStage`, no?

src/ReinforcementLearningCore/src/policies/agent/agent_srt_cache.jl
```diff
@@ -12,13 +12,37 @@ mutable struct SRT{S,R,T}
...
+end
+
+struct SART{S,A,R,T}
+    state::S
+    action::A
+    reward::R
+    terminal::T
+end
+
+function Base.push!(ts::Traces, xs::SA)
```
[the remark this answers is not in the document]
jeremiahpslewis: Hey! You're quite right that this is a limited work around. There are methods which should handle the MultiplexTrace already, but in this case (I believe / hope) we can get around the type-instable Trace/Trajectory indexing by hard-coding the indices. If it works, it's like a fast track which can be dropped when more abstract approaches also give acceptable performance.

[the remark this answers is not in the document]
jeremiahpslewis: I've specialized this for `CircularArraySARTTraces`, perhaps a safer approach?

src/ReinforcementLearningBase/src/base.jl
jeremiahpslewis: The `test_` functions are part of base and let you test your environments. So they aren't tests, but rather 'test mechanisms', which we use in the tests (but other people can / do as well).

src/ReinforcementLearningBase/src/interface.jl
jeremiahpslewis: Very good point 🙈

src/ReinforcementLearningCore/src/core/run.jl
jeremiahpslewis: 👍 Ok to save this for another PR?

jeremiahpslewis:
@HenriDeh two weeks later...tests pass! 🥳 Mind giving this a (final???) review?

src/ReinforcementLearningCore/src/policies/learners.jl
```diff
@@ -7,7 +7,8 @@ abstract type AbstractLearner end
 
 Base.show(io::IO, m::MIME"text/plain", L::AbstractLearner) = show(io, m, convert(AnnotatedStructTree, L))
 
-estimate_reward(L::AbstractLearner, env::AbstractEnv) = env |> state |> send_to_device(L) |> estimate_reward(L) |> send_to_device(env)
+# Take Learner and Environment, get state, send to estimate_reward(Learner, State)
+estimate_reward(L::Le, env::E) where {Le <: AbstractLearner, E <: AbstractEnv} = env |> state |> send_to_device(L.approximator) |> x -> RLCore.estimate_reward(L, x) |> send_to_device(env)
```
HenriDeh, line 11: Sorry I'm gonna bother you one more time. I hadn't seen it before but `estimate_reward` does not estimate rewards but the (possibly discounted) expected return. So I'd be in favor of renaming this `estimate_return` to avoid confusing future contributors. Additionnaly, estimate_reward might be a name we wish to keep for model-based reinforcement learning as reward prediction is a thing in e.g. Mu-Zero.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/mpo.jl
```diff
@@ -111,17 +111,17 @@ function update_critic!(p::MPOPolicy, batches)
         s, s′, a, r, t, = send_to_device(device(p.qnetwork1), batch)
         γ, τ = p.γ, p.τ
 
-        a′ = p.actor(p.rng, s′; is_sampling=true, is_return_log_prob=false)
+        a′ = RLCore.estimate_reward(p.actor, p.rng, s′; is_sampling=true, is_return_log_prob=false)
```
HenriDeh, line 114: I'm going to use MPO because I know the algorithm but this is valid for all actor critics. The other issue that I see with estimate reward is that now we use this function to call actor networks too. But this does not estimate a reward nor a return. It outputs an action. Therefore, `estimate_return` is not a better name. Concretely, this function only does a forward pass to the underlying model. So we could simply call it that? `forward`, like in pytorch?

jeremiahpslewis: Thanks for thinking about this...I was somewhat unsure about the proper naming. Do you want two methods, `estimate_return` and `forward` or should everything be `forward`?

jeremiahpslewis: The good news is that it's just a find-replace to fix it, but I would like to get the API naming (as close to) right as possible with this PR.

jeremiahpslewis: I'd love to merge this PR change (https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/884) into this PR, what are your thoughts?

HenriDeh: The easiest I think is to stay with `forward` for now. The distinction between estimate_return and estimate_policy could be useful, be it only for api clarity, but would require a distinction between value approximators and policy approximators. I think that's a bit too much for this already massive PR (119 files changed oof). Interesting for later though, as I'm kind of questioning the usefulness of Approximator in the the first place.

jeremiahpslewis:
One more review please! :) @HenriDeh

PR #889, opened 2023-05-24
Stacknamespace

HenriDeh:
I had a couple of issues when using the lastest master. First, there is now the `Base.stack` function that collides with Flux's. So I dropped an import in RLCore to avoid a precompilation error. 
I am wondering why this wasn't caught by previous PR's tests, this should interrogate us on whether the CI is correctly setup.
Second, the compats of RL.jl was not up to date with the version of the local subpackages.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


Project.toml
```diff
@@ -15,10 +15,10 @@ ReinforcementLearningZoo = "d607f57d-ee1e-4ba7-bcf2-7734c1e31854"
 [compat]
 Reexport = "0.2, 1"
 julia = "1.6"
-ReinforcementLearningBase = "0.11"
-ReinforcementLearningCore = "0.9"
-ReinforcementLearningEnvironments = "0.7"
-ReinforcementLearningZoo = "0.6"
+ReinforcementLearningBase = "0.12"
```
jeremiahpslewis, line 18: Do we really want to do this? At the moment, the sub-projects are under active development, but the overall package has been left at the older version sub-project compats. If we update these compats (which have breaking changes), we should also bump and release a breaking change-version of the overall RL.jl package.

But...there is still a fair amount of code in RLZoo, etc. which is part of the repo, but not `include`'ed into the packages. And basic stuff like q-learning doesn't work as we currently don't ship a tabularapproximator.

I would prefer to wait on updating these compats until all files in RLZoo are either migrated to the new syntax, tested and `include`'ed, as I think we're going to have to provide a fair amount of support to people when they start using the upgraded RL.jl package.

HenriDeh: You're right. I'll look into why I can't use my local dev version.

PR #897, opened 2023-06-02
NFQ

### LINKED ISSUE #895: Contribute Neural Fitted Q-iteration algorithm
Hi, I implemented Neural Fitted Q-iteration (NFQ) from [1] with the ReinforcementLearning.jl framework and would like to contribute to ReinforcementLearningZoo.jl. Currently it is in a separate project with version 0.10.2 of ReinforcementLearning.jl as dependency. 

I did not yet manage to integrate NFQ with the codebase here, as it seems to have changed quite a lot compared to the version available at the Julia package general registry. If anybody could provide some guidance with the pull request, I would be glad to contribute my implementation.

[1] Riedmiller, M. (2005). Neural Fitted Q Iteration – First Experiences with a Data Efficient Neural Reinforcement Learning Method. In: Gama, J., Camacho, R., Brazdil, P.B., Jorge, A.M., Torgo, L. (eds) Machine Learning: ECML 2005. ECML 2005. Lecture Notes in Computer Science(), vol 3720. Springer, Berlin, Heidelberg. https://doi.org/10.1007/11564096_32

CasBex:
Fixes #895 

PR Checklist
- [ ] Refactor old NFQ code
- [ ] Categorise NFQ
- [ ] Verify NFQ with example
- [ ] Code review


CasBex:
5e6f09d was functional in a separate julia project where ReinforcementLearning.jl was simply imported (for reference). Refactor c423048 is untested.

NFQ is meant to be used together with `QBasedPolicy`. During an episode, action selection happens exactly the same as in any Q-based policy, with the sole difference that the `optimise!` step should never happen during an episode, only after each episode / at the start of a new episode.

1. I saw no test code anywhere in the repo. Am I supposed to add tests to the repo or keep them separate? Is there an example somewhere that uses the refactored library?
2. I don't know what the appropriate location is since it's not quite a DQN algorithm but also not fully offline, feel free to move the files as appropriate.

src/ReinforcementLearningZoo/src/algorithms/offline_rl/NFQ.jl
```diff
@@ -0,0 +1,78 @@
+"""
+    NFQ{A<:AbstractApproximator, F, R} <: AbstractLearner
+    NFQ(approximator::A, num_iterations::Integer epochs::Integer, loss_function::F, batch_size::Integer, rng::R, γ::Float32) where {A<:AbstractApproximator, F, R}
+Neural Fitted Q-iteration as implemented in [1]
+
+# Keyword arguments
+- `approximator::AbstractApproximator` neural network
```
jeremiahpslewis, line 7:
```suggestion
- `approximator:: NeuralNetworkApproximator` neural network
```

src/ReinforcementLearningZoo/src/algorithms/offline_rl/NFQ.jl
```diff
@@ -0,0 +1,78 @@
...
+    x
+end
+
+function RLBase.plan!(learner::NFQ, env::AbstractEnv)
+    as = action_space(env)
+    return vcat(repeat(state(env), inner=(1, length(as))), transpose(as)) |> x -> send_to_device(device(learner.approximator), x) |> learner.approximator |> send_to_host |> vec
+end
+
+# Avoid optimisation in the middle of an episode
+function RLBase.optimise!(::NFQ, ::NamedTuple) end
```
jeremiahpslewis, line 50: Instead of doing 'type piracy', I think we should discuss whether the optimize step belongs after the `::PostEpisodeStage`, I actually stumbled upon this 'quirk' today myself. What do you think @HenriDeh

src/ReinforcementLearningZoo/src/algorithms/offline_rl/NFQ.jl
HenriDeh: I think `NeuralNetworkApproximator` does not exist anymore. It's only `Approximator` now as it can totally be a linear approximator as well.

src/ReinforcementLearningZoo/src/algorithms/offline_rl/NFQ.jl
HenriDeh: Why is it type piracy? The NFQ type and this function are defined in the Zoo package. 

However, I also think that the optimise is only called at the pre-act stage is a problem for ther many algorithms that optimise at the post-episode stage. We could extend the current implementation to `optimise!(usual_args..., ::AbstractStage)` and call it at both pre-act and post-episode in the run loop?

HenriDeh:
> 1. I saw no test code anywhere in the repo. Am I supposed to add tests to the repo or keep them separate? Is there an example somewhere that uses the refactored library?

Algorithms are tested by implementing a functioning example experiment in the RLExperiments package. This is a great way to document your algorithm also as people can see how you build a working agent with the correct trajectory. If you create some functions such as a loss, you can also add tests to see if they behave as expected. 

>I don't know what the appropriate location is since it's not quite a DQN algorithm but also not fully offline, feel free to move the files as appropriate.

I think the `dqns` directory is fine. 

> I am not very familiar with Flux so any comments on that side are welcome as well

I can review your implementation, just ping me when it is ready to be, or if you have a question.

src/ReinforcementLearningZoo/src/algorithms/offline_rl/NFQ.jl
Mytolo: I would also prefer to add the Stage to the optimise function and predefine optimse!(usual_args..., ::PostActStage) = nothing as a lot of algorithms train on whole episodes. I also thought to add a whole episode sampler for the trajectories.  (-> new issue for that?)

HenriDeh: >  I also thought to add a whole episode sampler.

Yes, I had the same thought, there's no way of doing that at the moment. 

Regarding optimise, however, I figured that we do not need to dispatch it on stages: it suffices to define the optimise! function like:
```
if is_terminated(env)
    #optimise etc.
end
```
and voila. Optimise is called right after `act!` so if the episode is over, we optimise, otherwise we go on.

jeremiahpslewis: @HenriDeh Type piracy was clearly the wrong term, good point. I think we should resolve this 'issue' before addressing this pull request, here's a PR where we can focus on the `optimise` behavior...I think the trajectory `push` logic may need to be tweaked as well...https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/899

Mytolo: In my opinion, an algorithm might profit from dispatching on different stages:

A message exchange policy for example may exchange messages in order to optimise agent's policies and after the episode they look on the whole episode for example. Also, if the Stage is included in the optimise! call, the compiler might profit from this information and compile to a narrower level than using if-else switches, right?

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,87 @@
...
+# Avoid optimisation in the middle of an episode
+function RLBase.optimise!(::NFQ, ::NamedTuple) end
+
+# Instead do optimisation at the end of an episode
+function Base.push!(agent::Agent{<:QBasedPolicy{<:NFQ}}, ::PostEpisodeStage, env::AbstractEnv)
+    for batch in agent.trajectory
+        _optimise!(agent.policy.learner, batch, env)
+    end
+end
+
```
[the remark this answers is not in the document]
CasBex: @HenriDeh @jeremiahpslewis the main branch seems to have the same issue in `RLCore.optimise!(::AbstractPolicy, ::S, ::Trajectory) where {S<:AbstractStage}`. `optimise!(policy, stage, batch)` is never called in the PostEpisode stage except at the very end of the simulation.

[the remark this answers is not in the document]
HenriDeh: Do you mean at the end of the Experiment ? Are you sure it is not due to the `threshold` or the `ratio` parameters of the Trajectory ?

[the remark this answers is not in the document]
CasBex: I am not too familiar with the Trajectory syntax. I've created a [separate branch](https://github.com/CasBex/ReinforcementLearning.jl/tree/debugNFQ) to show what I mean. If you test the NFQ example, only the `PostActStage` is actually called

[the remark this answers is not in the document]
HenriDeh: So the issue was that:
```
function RLBase.optimise!(policy::AbstractPolicy, stage::S, trajectory::Trajectory) where {S<:AbstractStage}
    for batch in trajectory
        optimise!(policy, stage, batch)
    end
end
``` 
in base.jl is dispatched on every stage and not only the stages where you want to optimise. This means that at every post-act-stage, your policy would sample the trajectory and call the dummy optimise of the stage. Once at the end of the episode, because the ratio of the controller is 1, the trajectory returns nothing when looped, because your already sampled it once per step (which is what the ratio = 1 wants). 

@jeremiahpslewis this is significant bug caught here. 

@CasBex I'm taking the liberty of pushing a fix to your branch. But I'll make a separate PR so that we don't have to wait for your NFQ to be merge for this fix to be available to the rest of us. I also took the liberty making a couple of quick tweaks to your code. 
It also meant I had to change the QBasedPolicy code.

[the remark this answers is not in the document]
CasBex: It's better now, but not quite there yet. With the fixes you pushed, the experiment runs, but the loop runs as many times as there are steps in the episode since the trajectories sample every step; I only need to call `optimise!` once. I've applied a quick fix in 2ed5ffb but this is obviously not ideal.

It may be a good idea to implement a "sample at the end of an episode" controller in `ReinforcementLearningTrajectories.jl`, but I propose to keep that outside the scope of this PR.

CasBex:
As far as I'm concerned, this PR is ready for review. The results of the test are attached. 

The provided example performs 500 random steps and then stops exploration. Contrary to the setup in the original paper [1], the example retrains NFQ at the end of every episode.

I noticed while tuning the example that NFQ is somewhat sensitive to the batch size. If the batch size is too small performance oscillates (but some episodes NFQ still obtains the maximum score). I think that could be caused by having 'bad luck' with the sample since one `optimise!` call can drastically change NFQ behavior. Larger batch size results in more consistent performance, so I chose the batch size equal to the amount of steps.

[1] Riedmiller, M. (2005). Neural Fitted Q Iteration – First Experiences with a Data Efficient Neural Reinforcement Learning Method. In: Gama, J., Camacho, R., Brazdil, P.B., Jorge, A.M., Torgo, L. (eds) Machine Learning: ECML 2005. ECML 2005. Lecture Notes in Computer Science(), vol 3720. Springer, Berlin, Heidelberg. https://doi.org/10.1007/11564096_32

JuliaRL_NFQ_CartPole

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,80 @@
...
+RLCore.forward(L::NFQ, s::AbstractArray) = RLCore.forward(L.approximator, s)
+
+function RLCore.forward(learner::NFQ, env::AbstractEnv)
+    as = action_space(env)
+    return vcat(repeat(state(env), inner=(1, length(as))), transpose(as)) |> x -> send_to_device(device(learner.approximator), x) |> x->RLCore.forward(learner, x) |> send_to_host |> vec 
+end
+
+# Avoid optimisation in the middle of an episode
+function RLBase.optimise!(::NFQ, ::AbstractStage, ::Trajectory) end
```
HenriDeh, line 56: Normally this is not necessary anymore with the latest master

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,80 @@
...
+    Q = learner.approximator
+    γ = learner.γ
+    loss_func = learner.loss_function
+    as = learner.action_space
+    las = length(as)
+    batch = nothing
+    for b in trajectory
+        batch = b
+    end
+    batch === nothing && return
```
HenriDeh, line 68: I see why you are doing this but it is not how a Trajectory is supposed to be used. Make a `for batch in trajectory` loop around the code below and it should be fine. Because here you are sampling potentially 200 times (for a cartpole episode) and using only the last batch. As you mentioned, because you want to sample between each episode, a `EpisodeSampleRatioController` might be needed. It should not be too hard to implement (just a InsertSampleRatioController that counts inserts only when terminal = true). 

If you don't want to implement such a controller in a separate PR I understand. In that case you case also replace this selected piece of code with `batch = sample(trajectory)` to bypass the controller. Note that by doing this you lose some features such as the threshold of the controller and the possibility to adjust the number of episodes between samplings.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,80 @@
...
+#     num_iterations::Integer = 20,
+#     epochs::Integer = 1000,
+#     loss_function::Any = mse,
+#     rng=Random.default_rng(),
+#     γ::Float32 = 0.9f0,
+#     ) where {A}
+#     NFQ(action_space, approximator, num_iterations, epochs, loss_function, rng, γ)
+# end
+
+# Copied from BasicDQN but sure whether it's appropriate
```
HenriDeh, line 45: Maybe drop these comments to make the source files cleaner.

src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
```diff
@@ -0,0 +1,95 @@
...
+    agent = Agent(
+        policy=QBasedPolicy(
+            learner=NFQ(
+                action_space=action_space(env),
+                approximator=Approximator(
+                    model=Chain(
+                        Dense(ns+na, 5, σ; init=glorot_uniform(rng)),
+                        Dense(5, 5, σ; init=glorot_uniform(rng)),
+                        Dense(5, 1; init=glorot_uniform(rng)),
+                    ) |> gpu,
```
HenriDeh, line 37:
```suggestion
                    ),
```
I dropped the gpu calls in #907 because it was only causing troubles and slowing down local testing, CI does not test on GPUs anyways.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,80 @@
...
+    
+    (s, a, r, ss) = batch[[:state, :action, :reward, :next_state]]
+    a = Float32.(a)
+    s, a, r, ss = map(x->send_to_device(device(Q), x), (s, a, r, ss))
+    for i = 1:learner.num_iterations
+        # Make an input x samples x |action space| array -- Q --> samples x |action space| -- max --> samples
+        G = r .+ γ .* (cat(repeat(ss, inner=(1, 1, las)), reshape(repeat(as, outer=(1, size(ss, 2))), (1, size(ss, 2), las)), dims=1) |> x -> maximum(RLCore.forward(Q, x), dims=3) |> vec)
+        for e = 1:learner.epochs
+            Flux.train!((x, y) -> loss_func(RLCore.forward(Q, x), y), params(Q.model), [(vcat(s, a), transpose(G))], Q.optimiser)
+        end
```
HenriDeh, line 78: do I understand correctly that after *each* episode, you will update the NN `epochs` times with the same target return, so `num_iterations*epochs` times with the same batch ? Sounds *very* offpolicy. Is there not supposed to be a new batch sampling at each iteration of the outer loop ? I'm not familiar with the algorithm, I'm just asking to be sure, it just looks very overfitting to a single batch.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
CasBex: 5ab7a1c

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
CasBex: That is not fully correct. Every iteration in the outer loop, the target is updated using the Bellman equation. Then the NN trains `epochs` times on the target. The estimated Q-function is then used to update the target again with the Bellman equation. The idea is much the same as value iteration in dynamic programming. 

By doing this only occasionally on a larger but fixed batch, higher data efficiency can be achieved. So yes, we are (over)fitting on a single batch, but this batch is intended to be large and representative of the actual environment dynamics, whereas with DQN the batches tend to be smaller (at least in the examples I saw in this repo).

More info can be found in Algorithms for Reinforcement Learning by Szepesvari https://sites.ualberta.ca/~szepesva/rlbook.html or the paper by Riedmiller which I cited earlier.

HenriDeh: Sure. Well, as long as it correctly reproduces the paper it's fine.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
CasBex: Resolved in 033dcdf. I decided to use the second method

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
HenriDeh: Right, NFQ is a learner not a policy. Then we should add 

`function RLBase.optimise!(::AbstractLearner, ::AbstractStage, ::Trajectory) end` 

in [learners.jl](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/src/policies/learners.jl). Mind doing that instead ? That way no unexpected behavior for future new learners.

src/ReinforcementLearningCore/src/policies/q_based_policy.jl
```diff
@@ -37,4 +37,4 @@ end
 RLBase.prob(p::QBasedPolicy{L,Ex}, env::AbstractEnv) where {L<:AbstractLearner,Ex<:AbstractExplorer} =
     prob(p.explorer, forward(p.learner, env), legal_action_space_mask(env))
 
-RLBase.optimise!(p::QBasedPolicy{L,Ex}, stage::S, x::NamedTuple) where {L<:AbstractLearner,Ex<:AbstractExplorer, S<:AbstractStage} = optimise!(p.learner, x)
+RLBase.optimise!(p::QBasedPolicy{L,Ex}, stage::S, t::Trajectory) where {L<:AbstractLearner,Ex<:AbstractExplorer, S<:AbstractStage} = optimise!(p.learner, stage, t)
```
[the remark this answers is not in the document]
HenriDeh: You are refering to the conflict right ? By default `QBasedPolicy` optimises at PreActStage now (I just noticed I mistakenly set PostActStage I'm gonna fix this right away). You can specialize (it's not piracy don't worry) this function to `RLBase.optimise!(p::QBasedPolicy{<:NFQ}, ::PostEpisodeStage, traj)` for your learner.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
CasBex: Resolved in 31b55b5 and b53c96b

CasBex: @HenriDeh  I see that you included the same fixes in #909. I'll wait till that is merged and then see about rolling back above commits and merging the updated main. After that NFQ should be good to go

HenriDeh: Okay, I'll wait for the review.

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
```diff
@@ -0,0 +1,66 @@
...
+function RLCore.forward(learner::NFQ, env::AbstractEnv)
+    as = action_space(env)
+    return vcat(repeat(state(env), inner=(1, length(as))), transpose(as)) |> x -> send_to_device(device(learner.approximator), x) |> x->RLCore.forward(learner, x) |> send_to_host |> vec 
+end
+
+function RLBase.optimise!(::NFQ, ::NamedTuple) end
+
+function RLBase.optimise!(p::QBasedPolicy{L,Ex}, s::PostEpisodeStage, trajectory::Trajectory) where {L<:NFQ,Ex<:AbstractExplorer} 
+       RLBase.optimise!(p.learner, s, trajectory)
+end
```
HenriDeh, line 46: This is no longer necessary, this is the current behavior of optimise at every stage for QBasedPolicy. Sorry for the recurring changes.

HenriDeh (commented):
Just that last little change then we're good to merge I think. Thanks for the PR, it helped us clean up a bit the learners shenanigans.

CasBex:
I found that the NFQ docs were also a bit outdated, so I'm updating that as well. Give me some time to get this in order because my git history is a bit messed up

src/ReinforcementLearningZoo/src/algorithms/dqns/NFQ.jl
CasBex: f77e198

CasBex:
@HenriDeh It should be done now, thanks for helping me with this

HenriDeh (dismissed):

HenriDeh:
I'm letting you merge in case you have a last minute tweak to do.

CasBex:
I don't have anything more to add, this branch is ready. I don't have write permissions to main, so I'll let you do the honors

HenriDeh (approved):

PR #899, opened 2023-06-05
Sketch out optimise! refactor

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -101,9 +101,7 @@ function _run(policy::AbstractPolicy,
             push!(policy, PostActStage(), env)
             push!(hook, PostActStage(), policy, env)
 
-            if is_terminated(env)
-                optimise!(policy)
-            end
+            optimise!(policy)
 
             if check_stop(stop_condition, policy, env)
                 is_stop = true
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -91,20 +91,20 @@ function _run(policy::AbstractPolicy,
         push!(policy, PreEpisodeStage(), env)
         optimise!(policy, PreActStage())
         push!(hook, PreEpisodeStage(), policy, env)
-        
+
 
         while !reset_condition(policy, env) # one episode
             push!(policy, PreActStage(), env)
             optimise!(policy, PreActStage())
             push!(hook, PreActStage(), policy, env)
-            
+
             action = RLBase.plan!(policy, env)
             act!(env, action)
 
             push!(policy, PostActStage(), env)
             optimise!(policy, PostActStage())
             push!(hook, PostActStage(), policy, env)
-            
+
             if check_stop(stop_condition, policy, env)
                 is_stop = true
                 push!(policy, PreActStage(), env)
@@ -118,7 +118,7 @@ function _run(policy::AbstractPolicy,
         push!(policy, PostEpisodeStage(), env)  # let the policy see the last observation
         optimise!(policy, PostEpisodeStage())
         push!(hook, PostEpisodeStage(), policy, env)
-        
+
     end
     push!(policy, PostExperimentStage(), env)
     push!(hook, PostExperimentStage(), policy, env)
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -101,7 +101,7 @@ function _run(policy::AbstractPolicy,
             push!(policy, PostActStage(), env)
             push!(hook, PostActStage(), policy, env)
 
-            optimise!(policy)
+            optimise!(policy, PostActStage())
 
             if check_stop(stop_condition, policy, env)
                 is_stop = true
@@ -114,6 +114,7 @@ function _run(policy::AbstractPolicy,
 
         push!(policy, PostEpisodeStage(), env)  # let the policy see the last observation
         push!(hook, PostEpisodeStage(), policy, env)
+        optimise!(policy, PostActStage())
     end
     push!(policy, PostExperimentStage(), env)
     push!(hook, PostExperimentStage(), policy, env)
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -105,7 +105,6 @@ function _run(policy::AbstractPolicy,
             optimise!(policy, PostActStage())
             push!(hook, PostActStage(), policy, env)
             
-
             if check_stop(stop_condition, policy, env)
                 is_stop = true
                 push!(policy, PreActStage(), env)
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -98,10 +98,10 @@ function _run(policy::AbstractPolicy,
             action = RLBase.plan!(policy, env)
             act!(env, action)
 
-            optimise!(policy)
-
             push!(policy, PostActStage(), env)
             push!(hook, PostActStage(), policy, env)
+
+            optimise!(policy)
 
             if check_stop(stop_condition, policy, env)
                 is_stop = true
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -101,7 +101,9 @@ function _run(policy::AbstractPolicy,
             push!(policy, PostActStage(), env)
             push!(hook, PostActStage(), policy, env)
 
-            optimise!(policy)
+            if is_terminated(env)
+                optimise!(policy)
+            end
 
             if check_stop(stop_condition, policy, env)
                 is_stop = true
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -89,23 +89,27 @@ function _run(policy::AbstractPolicy,
     while !is_stop
         reset!(env)
         push!(policy, PreEpisodeStage(), env)
+        optimise!(policy, PreActStage())
         push!(hook, PreEpisodeStage(), policy, env)
+        
 
         while !reset_condition(policy, env) # one episode
             push!(policy, PreActStage(), env)
+            optimise!(policy, PreActStage())
             push!(hook, PreActStage(), policy, env)
-
+            
             action = RLBase.plan!(policy, env)
             act!(env, action)
 
             push!(policy, PostActStage(), env)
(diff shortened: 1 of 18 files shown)


src/ReinforcementLearningCore/src/policies/agent/agent_srt_cache.jl
```diff
@@ -1,65 +1,49 @@
 import Base.push!
 
-mutable struct SRT{S,R,T}
+mutable struct SA{S,A}
```
Mytolo, line 3: Why is there now the action in it? I would let it like it is. Can you describe why you think that this is necessary? In my opinion, the reward and terminal information is necessary.

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -98,11 +98,13 @@ function _run(policy::AbstractPolicy,
             action = RLBase.plan!(policy, env)
             act!(env, action)
 
-            optimise!(policy)
-
             push!(policy, PostActStage(), env)
             push!(hook, PostActStage(), policy, env)
 
+            if is_terminated(env)
```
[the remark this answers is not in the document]
Mytolo: Also why is it put after PostActStage() calls? It should be directly after act!(), right?

src/ReinforcementLearningCore/src/policies/agent/agent_srt_cache.jl
Mytolo: Before anything is merged about the cache part and the handling of optimise, it should be thought of how the optimization may be done:

- Policy may be updated after each step agent takes
- Policy may only be updated after environment terminated
- The two above with different update semantics

src/ReinforcementLearningCore/src/policies/agent/multi_agent.jl
```diff
@@ -188,13 +188,24 @@ function Base.push!(multiagent::MultiAgentPolicy, ::PreActStage, env::E) where {
...
 
+# Multiagent Version
+function RLBase.plan!(agent::Agent{P,T,C}, env::E, p::Symbol) where {P,T,C,E<:AbstractEnv}
+    action = RLBase.plan!(agent.policy, env, p)
+    push!(agent.cache, action)
+    action
+end
+
 # Like in the single-agent case, push! at the PostActStage() calls push! on each player with the reward and termination status of the environment
 function Base.push!(multiagent::MultiAgentPolicy, ::PostActStage, env::E) where {E<:AbstractEnv}
```
Mytolo, line 199: multiagent[i].trajectory only exists, if the policies are of type Agent, right? In my opinion, the focus should be first on single agent and afterward thinking about multi-agent case.

Mytolo (commented):
As stated in the previous PR, it may be profitable - also mentioned in this review - to separate between optimise directly after the action execution **and / or** after the episode.
I welcome all of us to thoroughly think about these changes. They may incorporate a lot of refactoring of the other policies. 
Using dispatching for during and after episode will enable to use first one as default and allow policies to optimise directly after execution and / or after episode.

src/ReinforcementLearningCore/src/policies/agent/agent_srt_cache.jl
Mytolo: I do not get why we want to refactor the agent_cache here? The PR is about when to optimise. In my opinion, this issue (if it is actually one) has to be addressed in another PR.

src/ReinforcementLearningCore/src/policies/agent/multi_agent.jl
```diff
@@ -188,13 +188,24 @@ function Base.push!(multiagent::MultiAgentPolicy, ::PreActStage, env::E) where {
...
     for player in players(env)
-        push!(multiagent[player].cache, reward(env, player), is_terminated(env))
+        push!(agent.trajectory, multiagent[player].cache, reward(env, player), is_terminated(env))
```
Mytolo: When the environment is simultaneous, this is totally fine. 
If you think of sequential environments, this is never called, but the reward has to be collected by reward(env, player). Is this a problem or do we assume, that the multi_agent environment handles this?
Also, agent.trajectory has to be replaced by multiagent[player].trajectory in order to work.

src/ReinforcementLearningCore/src/policies/agent/multi_agent.jl
```diff
@@ -188,13 +188,24 @@ function Base.push!(multiagent::MultiAgentPolicy, ::PreActStage, env::E) where {
...
     end
 end
 
+function Base.push!(agent::Agent, ::PostExperimentStage, env::E, player::Symbol) where {E<:AbstractEnv}
```
Mytolo: Is this for cleaner garbage collection or why is this needed?

jeremiahpslewis:
I'm running into the following `optimise!` related issue with implementing basic q-learning:

We currently (main branch) do the following:
t=1, state, action, PUSH to traj, OPTIMISE, reward, t=2, state, action, PUSH, OPTIMISE reward...

But q-learning requires SARS, e.g. we need the state from t=2 to run optimization. Currently, we are optimizing with a 1-time-step lag, which just seems wrong (and I believe causes learning issues in some cases).

Ideally we would have 
t=1, state, OPTIMISE, action, PUSH to traj, reward, t=2, state, OPTIMISE, action, PUSH, reward...

for q-learning...

which seems to suggest we should run `optimise!`, `push!(hook` and  `push!(policy` at every stage, and implement the correct optimization position depending on the policy.

But...in general...I think we should `push!` before we optimise, as we otherwise are depriving the policy of the information which gets `push!` to the policy in that stage.

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -114,6 +114,7 @@ function _run(policy::AbstractPolicy,
 
         push!(policy, PostEpisodeStage(), env)  # let the policy see the last observation
         push!(hook, PostEpisodeStage(), policy, env)
+        optimise!(policy, PostActStage())
```
Mytolo, line 117: This should be
```suggestion
        optimise!(policy, PostEpisodeStage())
```
right?

src/ReinforcementLearningCore/src/core/stages.jl
```diff
@@ -20,3 +20,8 @@ Base.push!(p::AbstractPolicy, ::AbstractStage, ::AbstractEnv) = nothing
 Base.push!(p::AbstractPolicy, ::AbstractStage, ::AbstractEnv, ::Symbol) = nothing
 
 RLBase.optimise!(::AbstractPolicy) = nothing
+
+RLBase.optimise!(policy, ::PreEpisodeStage) = nothing
+RLBase.optimise!(policy::AbstractPolicy, ::PostEpisodeStage) = RLBase.optimise!(policy)
+RLBase.optimise!(policy, ::PreActStage) = nothing
+RLBase.optimise!(policy, ::PostActStage) = nothing
```
HenriDeh, line 27: This will introduce sneaky breaking changes to existing algorithms. I think by default all stages should be noop, or we keep it as is. If we change the default optimise stage in this PR, we must also fix the algorithms that optimise at each step.

HenriDeh: Also, I'm not sure why we would want to optimise at pre-act and pre-episode. But it doesn't hurt to have the calls I guess

HenriDeh: Ok nevermind, pre-act may be useful

Mytolo:
> I'm running into the following `optimise!` related issue with implementing basic q-learning:
> 
> We currently (main branch) do the following: t=1, state, action, PUSH to traj, OPTIMISE, reward, t=2, state, action, PUSH, OPTIMISE reward...
> 
> But q-learning requires SARS, e.g. we need the state from t=2 to run optimization. Currently, we are optimizing with a 1-time-step lag, which just seems wrong (and I believe causes learning issues in some cases).
> 
> Ideally we would have t=1, state, OPTIMISE, action, PUSH to traj, reward, t=2, state, OPTIMISE, action, PUSH, reward...
> 
> for q-learning...
> 
> which seems to suggest we should run `optimise!`, `push!(hook` and `push!(policy` at every stage, and implement the correct optimization position depending on the policy.
> 
> But...in general...I think we should `push!` before we optimise, as we otherwise are depriving the policy of the information which gets `push!` to the policy in that stage.


Well when using RLTrajectories, the next_state and state is only pushed to the data structure, once both are available. So you do nothing in the first optimise step if batching from trajectory. 
**BUT** i also think that it is more clearer to structure the loop as follows (this may result in the current optimise functions to be obsolete? @HenriDeh what do you think about?)

t = 0 
current state of environment 
plan action of agent
execute action
t = 1
reward of agent
next_state
push to trajectory
optimise

As long as the information is only pushed to the trajectory with complete information, this should not matter:
image
s: state
r: reward
t: time step
d: termination flag
orange: entities for optimization

src/ReinforcementLearningCore/src/core/stages.jl
HenriDeh: I think all should be defaulted to noop. Then, `optimise!(agent,...)` should dispatch on `agent.policy` + sample + stage. This dispatch is what should be overridden by the algorithm.

Mytolo: I also think that we should keep the default behavior to update after every step. Intuitively, this is also the most natural form of interaction. Adaption is done after every decision. 

Also, PreEpisode and PostEpisode should not be differentiated, as there is nothing added to infer an update for the policy.

jeremiahpslewis:
@Mytolo The problem I have is that `plan!`/`act!` is called for t=2 **before** `optimise!`, which means that we don't use t=1 information when deciding what to do.

Mytolo:
To be clear about this: You mean the initial state of the environment with **t = 1**?
This should not make much of a difference for a learning algorithm since the information is integrated afterwards, right?

jeremiahpslewis:
`plan!` is called for t=1 and t=2 with the same information state (as `optimise!` has yet to be called on a 'full' trajectory until *after* t=2 `plan!` / `act!`, so the first two steps are chosen 'randomly', but according to the algorithm, the second step should be chosen as a function of the explorer settings.

src/ReinforcementLearningCore/src/core/stages.jl
jeremiahpslewis: The idea is that you optimise on the full information provided by the policy + the policy `push!` which happens at each stage. As a result, every stage *potentially* has a different information set, which may or may not be the right one for triggering `optimise!`. But perhaps this is from an RL perspective nonsense, happy to adjust.

HenriDeh (approved):

jeremiahpslewis:
@Mytolo as I mentioned on slack, if you have any remaining comments, please let me know before tomorrow.

PR #900, opened 2023-06-08
fix bug in masked, multi-dimensional state space multi-threaded envs

Graeme22:
This fixes what appears to be a copy/paste bug that affects multi-threaded envs where the state space of the wrapped environment has different dimensionality than the action space of that environment.

For an example environment that this breaks, see https://github.com/Graeme22/rlskedge

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis (approved):

jeremiahpslewis:
@Graeme22 This looks like a good catch, thanks! Would you be able to contribute a test that would have caught the issue?

PR #907, opened 2023-06-15
Fix dispatch and update documentation

HenriDeh:
See #897 for the explanation.
I don't know how this affects current implementations of algorithms but it is likely to fix some rather than break anything.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh:
A consequence of this is that any algorithm must manually implement the `for batch in trajectory` loop as we cannot know at which stage the loop must live.

HenriDeh:
Tests should pass now. I will make some edits to the documentation before merging.

jeremiahpslewis (approved):

PR #909, opened 2023-06-19
QBasedPolicy optimise! forwards to learner. 

HenriDeh:
I mistakenly set the default update stage of QBasedPolicy to PostAct instead of PreAct in #907. This fixes it.

@jeremiahpslewis I thought of a perhaps better alternative. Change this function to simply dispatch the stage to its learner and define a `optimise!(learner, stage, trajectory)` for each learner. That way it is mandatory to explicitely decide the update stage when making a QBasedPolicy. 

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh:
> Wasn't the original `optimise!` call in the PostActStage? The [diff](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/899/files#diff-a79ee9cf7a42d0b59dd7620ed252683c6ffb8338f9c7deacb1a1b70e3d5e265e) of #899 seems to say this?

Indeed, so it wasn't my mistake. Should it stay PostAct ?

CasBex:
> > Wasn't the original `optimise!` call in the PostActStage? The [diff](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/899/files#diff-a79ee9cf7a42d0b59dd7620ed252683c6ffb8338f9c7deacb1a1b70e3d5e265e) of #899 seems to say this?
> 
> Indeed, so it wasn't my mistake. Should it stay PostAct ?

I have no particular preference. If you have a good reason to do it, then feel free.

HenriDeh:
I'd say no. The only difference is that PreAct is called an additional time for each episode, when the env is at its initial state.

HenriDeh:
I settled on keeping the PostActStage and also not having a default optimisation stage at all. My thinking is that with a default optimisation stage, an algorithm implementer must both define the new optimisation stage if it's not the default, and overload to remove the default one (which can lead to bugs if this is oversighted). 

I will add some documentation on that.

HenriDeh:
@jeremiahpslewis Do you have time for a quick review for this and the PR in Trajectories?

jeremiahpslewis:
This PR looks great! @HenriDeh

jeremiahpslewis (approved):

PR #919, opened 2023-07-04
couple of improvements to MPO

HenriDeh:
This simply 
- modifies slightly the math of vec_to_tril to create more stable covariance matrices 
- make the upperbound of the dual adaptive to let it be higher when rewards (and thus Q-values) have higher magnitudes.

### DIFF
--- a/src/ReinforcementLearningCore/src/utils/networks.jl
+++ b/src/ReinforcementLearningCore/src/utils/networks.jl
@@ -255,8 +255,9 @@ Transform a vector containing the non-zero elements of a lower triangular da x d
 function vec_to_tril(cholesky_vec, da)
     batch_size = size(cholesky_vec, 3)
     c2idx(i, j) = ((2da - j) * (j - 1)) ÷ 2 + i #return the position in cholesky_vec of the element of the triangular matrix at coordinates (i,j)
+    softplusbeta(x) = log(exp(0.1f0 * x) +1)*10f0 #a softer softplus to avoid vanishing values
     function f(j) #return a slice (da x 1 x batchsize) containing the jth columns of the lower triangular cholesky decomposition of the covariance
-        tc_diag = softplus.(cholesky_vec[c2idx(j, j):c2idx(j, j), :, :])
+        tc_diag = softplusbeta.(cholesky_vec[c2idx(j, j):c2idx(j, j), :, :]) .+ 1f-5
         tc_other = cholesky_vec[c2idx(j, j)+1:c2idx(j + 1, j + 1)-1, :, :]
         zs = ignore_derivatives() do
             zs = similar(cholesky_vec, da - size(tc_other, 1) - 1, 1, batch_size)
--- a/src/ReinforcementLearningCore/src/utils/networks.jl
+++ b/src/ReinforcementLearningCore/src/utils/networks.jl
@@ -250,23 +250,41 @@ function (model::CovGaussianNetwork)(state::AbstractMatrix, action::AbstractMatr
 end
 
 """
+    cholesky_matrix_to_vector_index(i, j)
+
+Return the position in a cholesky_vec (of length da) of the element of the lower triangular matrix at coordinates (i,j).
+
+For example if `cholesky_vec = [1,2,3,4,5,6]`, the corresponding lower triangular matrix is
+```
+L = [1 0 0
+     2 4 0
+     3 5 6]
+```
+and `cholesky_matrix_to_vector_index(3, 2) == 5`
+
+"""
+cholesky_matrix_to_vector_index(i, j, da) = ((2da - j) * (j - 1)) ÷ 2 + i
+softplusbeta(x, beta = 10f0) = log(exp(x/beta) +1)*beta #a softer softplus to avoid vanishing values
+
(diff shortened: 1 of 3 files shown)


HenriDeh:
@jeremiahpslewis can you validate please ?

src/ReinforcementLearningCore/src/utils/networks.jl
```diff
@@ -255,8 +255,9 @@ Transform a vector containing the non-zero elements of a lower triangular da x d
 function vec_to_tril(cholesky_vec, da)
     batch_size = size(cholesky_vec, 3)
     c2idx(i, j) = ((2da - j) * (j - 1)) ÷ 2 + i #return the position in cholesky_vec of the element of the triangular matrix at coordinates (i,j)
+    softplusbeta(x) = log(exp(0.1f0 * x) +1)*10f0 #a softer softplus to avoid vanishing values
```
jeremiahpslewis, line 258: Do we want to use a parameterized function which includes beta as an argument for softplusbeta? 

Something like `softplusbeta(x, β=100.f0) = 1f0 / β * log(1f0 + exp(β*x))` from this issue: https://github.com/FluxML/Flux.jl/issues/1216

Performance-wise there's no difference afaict.

src/ReinforcementLearningCore/src/utils/networks.jl
```diff
@@ -255,8 +255,9 @@ Transform a vector containing the non-zero elements of a lower triangular da x d
 function vec_to_tril(cholesky_vec, da)
```
jeremiahpslewis, line 255: JET finds lots of type issues with this function. Based on a scaffolding from ChatGPT, I came up with:

```julia
function vec_to_tril(v::AbstractVector{T}) where {T<:AbstractFloat}
    n = length(v)
    m = Int(floor(sqrt(2 * n + 1)))
    matrix = Matrix{T}(undef, m, m)

    # Fill the lower triangular part of the matrix with the elements of the vector
    index = 1
    for j = 1:m
        for i = j:m
            matrix[i, j] = i == j ? softplusbeta(v[index]) : v[index]
            index += 1
        end
    end

    return matrix
end
```

jeremiahpslewis: this is significantly faster and fewer allocations, but likely there are other (GPU???) downsides which are why this wasn't implemented originally. One difference (which could be fixed is that this function returns a matrix instead of a 3 dimensional array where the third dimension has a length of one.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/mpo.jl
```diff
@@ -224,7 +224,7 @@ end
 
 function solve_mpodual(Q::AbstractArray, ϵ)    
     g(η) = η * ϵ + η * mean(logsumexp( Q ./η .- Float32(log(size(Q, 2))), dims = 2))
-    Optim.minimizer(optimize(g, eps(ϵ), 10f0))
+    Optim.minimizer(optimize(g, eps(ϵ), maximum(abs.(Q))))
```
jeremiahpslewis, line 227: I don't have any idea what this is doing, so I'm going to trust you blindly on this one. :)

jeremiahpslewis:
@HenriDeh I looked at this PR and did my best to review it based on things I felt I can contribute. In terms of computational correctness, I don't really have a clue for these functions, but I would suggest that you add unit tests for the functions (if only so that if future changes to the code lead to different results, we can investigate and learn whether we've broken something or it was broken all along...

jeremiahpslewis (changes requested):
see above

src/ReinforcementLearningCore/src/utils/networks.jl
HenriDeh: Sadly this implementation is both Zygote incompatible due to array mutation and GPU incompatible due to scalar indexing. Coming up with a function that satisfies both of these requirements is the reason behind the weird design of the function in its current state. I think the Enzyme AD will solve the mutation problem and then it might be possible to also make a better GPU kernel.

src/ReinforcementLearningZoo/src/algorithms/policy_gradient/mpo.jl
HenriDeh: 10 was an arbitrary upperbound I took from another implementation. But in practice I observed that it depends on the scale of the elements in Q. I guess the former implementer worked with smaller scales than I do.

src/ReinforcementLearningCore/src/utils/networks.jl
jeremiahpslewis: Figures. What do you think about pulling out the three functions that are defined within `vec_to_tril` so it's easier to read? It'd also be easier to test them, though I don't think that's strictly necessary.

HenriDeh:
All tests are passing on my machine. The failing test in [e004bef](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/919/commits/e004bef4407363cee1f80af2671c8349b979e41f) seems to be a bug (a timer timed 4x10^233), quite impossible. I couldn't reproduce on my machine and it is unrelated to the changes of the PR. Hopefully this does not happen again in the CI job.

PR #921, opened 2023-07-07
Rework the run loop

HenriDeh:
This PR attempts to rework the run loop in order to fix data storage misalignment. This relies on the RLTrajectories 0.2 release that stores the data along with meta information about each step and keeps the traces aligned. 

The loop is now as follows
- PreEpisodeStage: the state is pushed alone in the trajectory.
- PreActStage: nothing
- plan! simply query and the action for the current state
- act!
- PostActStage: push the action, reward, terminal flag, and the new state to the trajectory. 
- check_stop: now only breaks the episode loop prematurely.
- loop to PreActStage unless reset condition
- PostEpisodeStage, checks for missing traces in the trajectory, typically :next_action. Only query for an action if :next_action is a name in the trajectory's keys. Therefore the user must make sure to use the correct trajectory. 

You'll notice that agent no longer has a cache. This is because reward, terminal and next_state can all be obtained at the PostActStage and immediately pushed to the trajectory. The action is given as an argument to push.

I also renamed agent/base.jl to agent/agent_base.jl because it was bugging me to have a base.jl tab that was unrelated to RLBase. 

PR Checklist

- [ ] Update NEWS.md?
- [ ] MultiAgent
- [ ] PR Review?


HenriDeh:
@Mytolo have a look. It's WIP but it gives an idea of what this could be. Feel free to make reviews.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -13,22 +13,21 @@ is a Callable and its call method accepts varargs and keyword arguments to be
...
 
         if TrajectoryStyle(trajectory) === AsyncTrajectoryStyle()
             bind(trajectory, @spawn(optimise!(policy, trajectory)))
         end
         agent
     end
 
-    function Agent(policy::P, trajectory::T, cache::C) where {P,T,C}
-        agent = new{P,T,C}(policy, trajectory, cache)
+    function Agent(policy::P, trajectory::T) where {P,T,C}
```
Mytolo: C is not needed any more? Or why did you keep it here?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -51,39 +49,37 @@ function RLBase.optimise!(policy::AbstractPolicy, stage::AbstractStage, trajecto
...
-    action
+function RLBase.plan!(agent::Agent, env::AbstractEnv, p::Symbol)
+    RLBase.plan!(agent.policy, env, p)
 end
 
-function Base.push!(agent::Agent{P,T,C}, ::PostActStage, env::E) where {P,T,C,E<:AbstractEnv}
-    push!(agent.cache, reward(env), is_terminated(env))
+function Base.push!(agent::Agent, ::PostActStage, env::AbstractEnv, action)
+    next_state = state(env)
+    push!(agent.trajectory, (state = next_state, action = action, reward = reward(env), terminal = is_terminated(env)))
```
Mytolo, line 70: Maybe I got sth wrong, but should next_state not stored in next_state field of the trajectory? next_state is successor of the state before the action was done in the environment, right?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -51,39 +49,37 @@ function RLBase.optimise!(policy::AbstractPolicy, stage::AbstractStage, trajecto
...
 end
 
-function Base.push!(agent::Agent, ::PostExperimentStage, env::E) where {E<:AbstractEnv}
-    RLBase.reset!(agent.cache)
-end
-
-function Base.push!(agent::Agent, ::PostExperimentStage, env::E, player::Symbol) where {E<:AbstractEnv}
-    RLBase.reset!(agent.cache)
+function Base.push!(agent::Agent, ::PostEpisodeStage, env::E)
+    if haskey(agent.trajectory, :next_action)
```
Mytolo, line 74: This is if the episode finished (whether truncated or terminated) we query the policy to plan another step. We should also check if the environment is not terminated? If it is, it just makes no sense to plan an action.

Mytolo (changes requested):
For the multi_agent_policy, the run loop has also to be adjusted (In sequential environment dispatch) I am also not sure if the way the experience is stored that way it's like i suppose it to be and would like to discuss it here:

Tuples in experience replay:

(s_0, a_0, r_1, s_1, d_1)
(s_1, a_1, r_2, s_2, d_2)
...
(s_t, a_t, r_t+1, s_t+1, d_t+1)

and so on. Is it like that? And for trajectories with next_action, there should also be a_t+1 which is different to that in the next state choosen one? Or is it the same? (Or does it even matter if it is or not?)

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -17,7 +17,7 @@ mutable struct Agent{P,T} <: AbstractPolicy
     policy::P
     trajectory::T
 
-    function Agent(policy::P, trajectory::T) where {P,T}
+    function Agent(policy::AbstractPolicy, trajectory::Trajectory)
```
Mytolo, line 20: Is this more efficient regarding typing? Or is it simply more readable?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
HenriDeh: It's an oversight. Already fixed.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
HenriDeh: It wouldn't make sense indeed, but if your environment has terminal states at all, then you should not use a trajectory that has a next_action key. That's the how I thought about it. If we add that check, then it allows the user to have an incorrect trajectory without an error being thrown and the buffer will accumulate mistakes.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
HenriDeh: `function Agent(policy::P, trajectory::T) where {P,T} ` was equivalent to `function Agent(policy::Any, trajectory::Any)`, or to `function Agent(policy, trajectory)`. 

Agent expects an abstract policy and a trajectory so it's more robust to user mistakes. Efficiencywise, I reckon it makes no difference at all.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
HenriDeh: It's the same. Both names point to the same Trace in the trajectory.

HenriDeh:
> For the multi_agent_policy, the run loop has also to be adjusted (In sequential environment dispatch)

Haven't looked at that yet but I should not forget thank you.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
HenriDeh: Nevermind, P and T were needed for the new{P,T} constructor below.

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
Mytolo: Right. It is the multiplex trace, right?

HenriDeh: Yes

docs/src/Zoo_Algorithms/MPO.md
```diff
@@ -56,7 +56,7 @@ The next step is to wrap this policy into an `Agent`. An agent is a combination
 
 ```julia
 trajectory = Trajectory(
-            CircularArraySARTTraces(capacity = 1000, state = Float32 => (4,),action = Float32 => (1,)), 
+            CircularArraySARTSTraces(capacity = 1000, state = Float32 => (4,),action = Float32 => (1,)),
```
jeremiahpslewis, line 59:
```suggestion
            CircularArraySARTSTraces(capacity = 1000, state = Float32 => (4,), action = Float32 => (1,)), 
```

src/ReinforcementLearningCore/src/core/run.jl
```diff
@@ -102,21 +102,17 @@ function _run(policy::AbstractPolicy,
             action = @timeit_debug timer "plan!"                RLBase.plan!(policy, env)
             @timeit_debug timer "act!"                          act!(env, action)
 
-            @timeit_debug timer "push!(policy) PostActStage"    push!(policy, PostActStage(), env)
+            @timeit_debug timer "push!(policy) PostActStage"    push!(policy, PostActStage(), env, action) #push a, r, t, s' to traj
```
jeremiahpslewis, line 105: These comments are helpful for SART, but are they generic to all RL experiments?

src/ReinforcementLearningCore/src/core/stages.jl
```diff
@@ -17,7 +17,9 @@ struct PreActStage <: AbstractStage end
 struct PostActStage <: AbstractStage end
 
 Base.push!(p::AbstractPolicy, ::AbstractStage, ::AbstractEnv) = nothing
+Base.push!(p::AbstractPolicy, ::PostActStage, ::AbstractEnv, action) = nothing
 Base.push!(p::AbstractPolicy, ::AbstractStage, ::AbstractEnv, ::Symbol) = nothing
+Base.push!(p::AbstractPolicy, ::PostActStage, ::AbstractEnv, action, ::Symbol) = nothing
```
jeremiahpslewis, line 22: This is begging for us to create an `action` type, but that's something for another PR. :)

jeremiahpslewis:
@HenriDeh This looks great! I'll run a couple of type stability checks after this gets closer to ✅

src/ReinforcementLearningCore/Project.toml
```diff
@@ -40,7 +40,7 @@ Parsers = "2"
 ProgressMeter = "1"
 Reexport = "1"
 ReinforcementLearningBase = "0.12"
-ReinforcementLearningTrajectories = "^0.1.9"
+ReinforcementLearningTrajectories = "^0.2.1"
```
jeremiahpslewis, line 43:
```suggestion
ReinforcementLearningTrajectories = "0.3"
```

PR #922, opened 2023-07-07
added first Independent Q Learning experiment

Mytolo:
PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?
Here is an experiment for RLExperiments showing how to set up a simple independent learning with two players. (In the example, the roles of randomly choosing an action and taking create_policy -> DQN can be changed and they can also both try to approximate their Q values for action selection)


src/ReinforcementLearningExperiments/src/experiments/MARL/IDQN_TicTacToe.jl
```diff
@@ -0,0 +1,96 @@
...
+using Flux: glorot_uniform
+
+using ProgressMeter
+
+
+rng = StableRNG(1234)
+
+cap = 100
+
+RLCore.forward(L::DQNLearner, state::A) where {A <: Real} = RLCore.forward(L, [state])
```
Mytolo, line 26: **NOTE:** This should be included into RLCore. For environments being identified by Reals, there should be a forward function for Approximators wrapping them into vectors s.t. they can be processed by neural networks.

Mytolo:
@HenriDeh, @jeremiahpslewis would you have a look? It is supposed to just be an example on how to use `multiAgentPolicy`.

HenriDeh (changes requested):
I don't see a problem with the implementation but it needs to be added to the tests of RLExperiments like the other ones are. Otherwise we may not see that a change breaks this example and it will live there broken until someone notices.

Mytolo:
> I don't see a problem with the implementation but it needs to be added to the tests of RLExperiments like the other ones are. Otherwise we may not see that a change breaks this example and it will live there broken until someone notices.

I this what you intended to say here @HenriDeh ? Commit [82f77f4](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/922/commits/82f77f49fe5607ff4ba78753f9dedf093af4e6e3)

src/ReinforcementLearningExperiments/test/runtests.jl
```diff
@@ -1,6 +1,14 @@
 using ReinforcementLearningExperiments
 using CUDA
 
+using Requires
+
+
+
+using Requires
```
HenriDeh, line 8: There's a duplicate here

HenriDeh (commented):
Yes thank you. Gotta figure out why the CI fails now.

src/ReinforcementLearningExperiments/test/runtests.jl
Mytolo, line 8:
```suggestion
```

src/ReinforcementLearningExperiments/test/runtests.jl
```diff
@@ -15,6 +23,8 @@ run(E`JuliaRL_VPG_CartPole`)
 run(E`JuliaRL_MPODiscrete_CartPole`)
 run(E`JuliaRL_MPOContinuous_CartPole`)
 run(E`JuliaRL_MPOCovariance_CartPole`)
+run(E`JuliaRL_IDQN_TicTacToe`)
+@require PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0" run(E`JuliaRL_DQN_MPESimple`)
```
Mytolo, line 27: this one is cause of failure. We want to run it if PyCall is present and pettingzoo is installed in python env. Will asap push a suggestion for this.

Mytolo:
Finally, the pipeline should be correct. I needed to add some parts, but this is also helpful for other 3rd party environments once they need to be tested.

Mytolo:
> I don't see a problem with the implementation but it needs to be added to the tests of RLExperiments like the other ones are. Otherwise we may not see that a change breaks this example and it will live there broken until someone notices.

this should be solved now @HenriDeh ?

HenriDeh:
Hi @Mytolo, sorry for the delay, I was on vacations. You have conflicts to resolve then I can approve the PR for merging :)

Issue #913, opened 2023-06-27
Executing RLBase.plan! after end of experiment


Mytolo:
Why do we execute RLBase.plan! **after the experiment is done**? Also we should test if the environment is terminated? 

For environments with FULL_ACTION_SET Action_style, this yields **an error as the legal_action_mask is empty**. Or the policies should be updated s.t. they do not return anything when environment is terminated for plan!? BUT this behavior would also be unexpected.

src\ReinforcementLearningCore\src\policies\agent\multi_agent.jl:133:140
```
if check_stop(stop_condition, policy, env)
       is_stop = true
       push!(multiagent_policy, PreActStage(), env)
       optimise!(multiagent_policy, PreActStage())
       push!(multiagent_hook, PreActStage(), policy, env)
       RLBase.plan!(multiagent_policy, env)  # let the policy see the last observation
       break
end
```

In my opinion, this should be completely omitted: Only do 
```
is_stop = true
break
```

HenriDeh:
I keep writing answers to this but it's a little confusing. 

Imagine the stop_condition is after 1 step. The run loop would be
```
PreAct: s0
plan!: a0 
<s0, a0> are in the trajectory
optimise!: nothing, we don't have a full step yet
PostAct: r0, t0
check_stop: true, we did 1 step
PreAct: s1
optimise!: nothing because we don't have a full step yet (did not call plan! to push r0 and t0)
plan!: a1
<s0,a0,r0,t0,s1,a1> are in the trajectory
```
It indeed makes no sense as we did not optimise even though we completed a step. But deleting the content of the if block would also prevent that optimization to happen, if it was setup correctly. 
Maybe we can move this the the PostEpisodeStage? This would not solve your issue. 
But we must push the last step in the cache. Maybe the trajectory push should not happen at plan!, to decouple the cache pushing at the action query ? Thoughts @jeremiahpslewis  ?

Mytolo:
In my opinion, this should be included in the normal run loop without the need to take care after the stop_condition is reached. It should be added right after the action was passed to the environment. push!(policy, PostActStage(), env) would be the perfect place for that, right?

Especially for sequential multi agent environment this off-sync (out of sync with environment steps and run loop step) causes a big problem when thinking about trajectory similarity. I recognized for pettingzoo which i included in PR #782:
Although there are environments with totally same reward at end of each agents state, the trajectories include different (shifted) rewards.

[...]

PR #926, opened 2023-07-11
refactored multi agent proposal

Mytolo:
Instead of returning a generator for RLBase.plan! use NamedTuples s.t. the environment is able to gather the actions for the players correctly. It might be that the iteration of the environment is not in order (for example, this might occur when it is wrapped from another language into julia).

Moreover, this PR includes proper push operations for the Basic `Agent` policy for simultaneous dynamicStyle trait envs. They were missing. Hence, the trajectory was not updated acordingly.


Mytolo:
NOTE: CI fail depends on rework of run loop as RLBase.plan! is executed after termination of environment as stated in #913, PR #921

HenriDeh:
Doesn't the new loop work for you ?

PR #932, opened 2023-07-21
fixed devmode / project files

Mytolo:
PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?

This re-enables the opportunity to use devmode.jl and activate_devmode!(). Needed to adjust compat for Project.toml file in main project. **NOTE:** This is not directly done, but have to be manually copied from Project.dev.toml to Project.toml in the root directory!

For devmode.jl, the dependency on ReinforcementLearning is not needed, as there should be as few dependencies as needed. Depending on the whole RL package might in some cases create cyclic dependencies. (IF one wants to include RLExperiments to RL as dependency, but this is just loosely.)

### DIFF
--- a/Project.dev.toml
+++ b/Project.dev.toml
@@ -1,0 +1,27 @@
+name = "ReinforcementLearning"
+uuid = "158674fc-8238-5cab-b5ba-03dfc80d1318"
+authors = ["Johanni Brea <jbrea@users.noreply.github.com>", "Jun Tian <tianjun.cpp@gmail.com>"]
+version = "0.11.0-dev"
+
+[deps]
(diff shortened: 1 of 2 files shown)


Mytolo:
What do you think about this @HenriDeh, @jeremiahpslewis ?

Mytolo:
> What about having a folder ‚dev/‚ putting this file as Project.toml in it and then all you have to do is to activate the dev project.

Well this would be another way, but then there has to be the devmode.jl in another folder and the paths have to be adjusted. But if you would prefer it this way, it would also be possible In my opinion the proposed file is cleaner.

PR #933, opened 2023-07-24
fixed DQNLearner Gpu isse

Mytolo:
PR Checklist

- [x] Update NEWS.md?
- [ ] PR Review?

In the DQNLearner during optimise, the batches were not copied to the suitable device. This caused errors when using the approximator on GPU. Also, as `'` is also reserved as operator in Julia, the naming convention `_next` is more clear and does not create weird errors due to being an operator.


Mytolo:
Actually, there should be basic experiments using GPU for all policies such that this does not happen any more.

jeremiahpslewis (approved):

PR #938, opened 2023-07-30
Reduce find_all_max allocations and increase speed based on chatgpt s…

jeremiahpslewis:
…uggestion

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh (dismissed):
Mindblowing to me that passing 3 times on the array is more efficient than twice just because it avoids allocations. And that the Base function findall is not efficient in the first place.

HenriDeh:
Actually, the scalar indexing might break this for GPU implementations. Might need to fall back to the current implementation for `x::CuArray`.

jeremiahpslewis:
> Actually, the scalar indexing might break this for GPU implementations. Might need to fall back to the current implementation for `x::CuArray`.

Great point, `GPUArrays` should do the trick for an appropriate fallback. Will add.

HenriDeh (approved):

PR #941, opened 2023-08-01
GPU: Drop to_device Logic and use Flux cpu / gpu functions

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis:
@findmyway Looked into this a bit further, I think that it would be idea if we can eliminate GPU handling logic (as far as possible) from this package and lean on `GPUArrays` and `Flux` which already has done extensive work to make things GPU agnostic. One piece which isn't yet elegantly handled in this PR is the GPU rng stuff, which can / should be made GPU agnostic. But before I do that, I'd love your thoughts on the general direction of this PR...

PR #948, opened 2023-08-05
Drop to_device, use Flux.gpu / Flux.cpu

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


src/ReinforcementLearningCore/test/runtests.jl
```diff
@@ -4,9 +4,22 @@ using ReinforcementLearningEnvironments
...
+    set_preferences!(flux_uuid, "gpu_backend" => "Metal")
+
+    using Metal
+else
+    using CUDA, cuDNN
+end
+
 using CircularArrayBuffers
 using Flux
+println("Flux.GPU_BACKEND = $(Flux.GPU_BACKEND)")
```
jeremiahpslewis, line 22:
```suggestion

println("Flux.GPU_BACKEND = $(Flux.GPU_BACKEND)")
```

HenriDeh:
Hi, I'm not against this change but I think it should be noted that this is technically breaking (probably for nobody) as this removes the ability to use multiple GPUs. Currently you can host your NNs on separate cards if you wish. I think this is okay though as large-scale RL is not really in the scope of this package. Thanks for the work, I see you've been busy.

jeremiahpslewis:
Agree about the breaking comment...my thoughts were that if we want to add this in the future, we could have code in a separate package like RLDistributed which overloads the relevant functions.

PR #959, opened 2023-08-11
type Experiment

### LINKED ISSUE #881: Fix Experiment type setup
See https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/879#discussion_r1185818098

HenriDeh:
Just to close #881

### DIFF
--- a/src/ReinforcementLearningCore/src/core/run.jl
+++ b/src/ReinforcementLearningCore/src/core/run.jl
@@ -33,11 +33,11 @@ function try_parse_kw(s)
     NamedTuple(kw)
 end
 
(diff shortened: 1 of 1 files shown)


jeremiahpslewis:
@HenriDeh You probably know this already, but this will take some porting of the `Experiement()` function / constructor so that RLExperiments works.

HenriDeh:
I tried to take a closer look at this and it is way too involved with the parsing stuff to be worth the time, given that the improvement we will get from typing Experiment is likely null. Closing this. @jeremiahpslewis are you okay with closing #881 ?

Issue #961, opened 2023-08-11
Missing features in RLCore


HenriDeh:
Giving the discussion in #960, I'd make a list of components of RLCore that I feel are missing or incomplete. 

- [ ] Target networks: they are currently named TwinNetwork. This is incorrect, twin_networks refer to another concept in RL. They would wrap an approximator, hold a target version of it, and handle the updating of both in one optimise! method.
- [ ] Double Q Network, aka the actual "twin networks" (as in TD3). Currently they must be implemented manually in the algorithms. They are typically optional though. So for the sake of experimenting, it would be nice to have a wrapper around a q-network approximator to automatically make a double of it and manage forwarding and training automatically. 
Ideally, DQN could wrap a standard Approximator or a TargetNetwork. In the latter case, it would encapsulate 4 NN in one struct, all of that seamlessly.
- [ ] State-Value-approximator and Action-Value-Approximator. Specialized versions of Approximators that would handle the common stuff happening when training these common types of networks. They would provide an interface to the user to, for example, overload the loss function (see the old #615). 
- [ ] A clean collection of Bellman Target Operators for value approximator updates: n-step TD, GAE, Retrace, Vtrace. Complemented with documentation on their usage (which trajectory and sampler to use). 
- [ ] A clear distinction between Policies and Learners. It occurred to me that DQN-based use a QBasedPolicy(::Learner) because each algorithm learns differently, but use the learner q_network in the same way. On the other hand, policy gradient algorithms use a custom `XXXPolicy` struct and that's it. The ActorCritic struct is a candidate for some policy gradient algorithms (eg. PPO, MPO, SAC) but there should be an underlying Actor and a Critic object. Currently the struct is not very general. 

I'm kind of using this issue as a notepad but I figured it'd be nice to share this.

PR #966, opened 2023-08-17
TargetNetwork

HenriDeh:
This PR started when I wanted to clean up a bit the RLCore structure. Then I noticed that TargetNetworks are currently named TwinNetworks (incorrect naming see #961). I went down the rabbit hole of refactoring it in addition to simply renaming it. I created a doc page, I invite you to read it to understand the point of this PR. 

I've refactored all DQNs algorithms, they are now agnostic to the use of a target or not. MPO also uses the new struct (here it is enforced by type restrictions).

I have yet to create some test for the new struct but I'd like your opinion on what I converged to.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh:
@jeremiahpslewis I tried to replicate the error of the experiments in [a8acb2d](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/966/commits/a8acb2d80d9761e38d36ad4bc07dfadaf0d71726). This seems to be an error that happens with very low probability. 

It looks like an action "0" is occasionally sampled an leads to an out of bounds error. This 0 I think is a dummy action used in RLTrajectories padding... So there's a sneaky bug to find.

jeremiahpslewis (commented):
@HenriDeh I can review things again once I've submitted my master's thesis end of next week. Looking forward to digging into this!

jeremiahpslewis (approved):

PR #968, opened 2023-08-24
Fix multi thread env state in case N=1

johannes-fischer:
I have come across an issue with `MultiThreadEnv` for an environment where the `state_space(envs[1]=` is not a `Space`. The error message is

ERROR: DimensionMismatch: cannot broadcast array to have fewer non-singleton dimensions

because `selectdim(menv.states, N, i)` in this case returned a
`0-dimensional view(::Vector{Vector{Float64}}, 1) with eltype Vector{Float64}`.

I belief this change should fix the problem.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


jeremiahpslewis:
Thanks for catching this! Can you add a test which catches this issue?

johannes-fischer:
Is there an environment where the state space is not a `Space`? The one I am using is not suitable for a test. Unfortunately, there seem to be no other tests to adapt.

jeremiahpslewis:
Can you create a minimal custom env extending AbstractEnv?

PR #972, opened 2023-09-08
add retrace

HenriDeh:
This PR adds a function to compute the retrace bellman operator as described in [this paper](https://paperswithcode.com/paper/safe-and-efficient-off-policy-reinforcement). It is described as an algorithm so I put it in RLZoo but it can be argued that this is more of a component used by RL algorithms . I'm thus open to move this to RLCore. 

The tests are incomplete. For now I only test if the function returns the expected operators given toy networks. This needs to also be tested with actual networks for the API. 

PR #966 must also be merged because this uses the new 'target' api.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh:
I'm closing this as I am not planning on finishing this soon, and the upcoming breaking changes mean one might as well start from scratch to implement this.

PR #974, opened 2023-09-13
Prioritised DQN GPU

### LINKED ISSUE #973: Prioritised DQN failing on GPU
With #971 coming to a close in `RLTrajectories.jl` I have identified another issue for the Prioritised DQN implementation: states/actions are not forwarded to the device of the Q function, and thus the gpu implementation is broken. I will work on a fix

CasBex:
Fixes #973. (Has been tested with `CUDA.allowscalar(false)`)

Two remarks
1. Currently, k and p' are returned as CuArrays. This does not seem to hamper the algorithm so I've not reverted them back, but if that is an issue it can be changed.
2. I have not included tests in the form of an experiment but simply modified the existing `JuliaRL_PrioritizedDQN_CartPole`. I can add that to the experiments but I think that it will just clutter the RLExperiments package.


HenriDeh:
Can't do simpler than that. If it works then let's keep the send_to_device for all traces.

HenriDeh (approved):

PR #980, opened 2023-09-25
Nfq refactor

CasBex:
Hi, I implemented NFQ a while ago (#897) and have since noticed some things that could be improved. Changes are summarized below:

1. Use a (state) -> ... -> |action space| structure for the Q-network (rather than (state, single action) -> ... -> value structure and looping over all actions)
2. Explicitly use gradient / optimise instead of Flux.train
3. Rely on the Trajectory controller for sampling rather than sampling inside the optimise! function

In addition to increased uniformity, this also consumes much less memory on the gpu compared to the previous version, which needed to duplicate all input data |action space| times to loop over all actions. The main drawback is that it is less faithful to the original implementation [1].

[1] Riedmiller, M. (2005). Neural Fitted Q Iteration – First Experiences with a Data Efficient Neural Reinforcement Learning Method. In: Gama, J., Camacho, R., Brazdil, P.B., Jorge, A.M., Torgo, L. (eds) Machine Learning: ECML 2005. ECML 2005. Lecture Notes in Computer Science(), vol 3720. Springer, Berlin, Heidelberg. https://doi.org/10.1007/11564096_32

### DIFF
--- a/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
+++ b/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
@@ -56,7 +56,7 @@ function RLCore.Experiment(
                 state=Float32 => (ns,),
             ),
             sampler=BatchSampler{SS′ART}(
-                batch_size=2048,
+                batch_size=128,
                 rng=rng
             ),
             controller=InsertSampleRatioController(
--- a/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
+++ b/src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
@@ -21,31 +21,32 @@ function RLCore.Experiment(
     ::Val{:CartPole},
     seed = 123,
 )
+
     rng = StableRNG(seed)
     env = CartPoleEnv(; T=Float32, rng=rng)
-    ns, na = length(state(env)), length(first(action_space(env)))
+    ns, na = length(state(env)), length(action_space(env))
 
     agent = Agent(
         policy=QBasedPolicy(
             learner=NFQ(
-                action_space=action_space(env),
                 approximator=Approximator(
                     model=Chain(
-                        Dense(ns+na, 5, σ; init=glorot_uniform(rng)),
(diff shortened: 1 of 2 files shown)


CasBex:
Unrelated, but during testing I tried to work with the EpisodeSampleRatioController and didn't manage to get it working: it complains that `NamedTuple has no field terminal`. I didn't find any examples of it in this repo or in RLTrajectories, so that may be worthwile to pursue in another issue.

src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
```diff
@@ -21,54 +21,57 @@ function RLCore.Experiment(
...
         ),
         trajectory=Trajectory(
             container=CircularArraySARTSTraces(
                 capacity=10_000,
                 state=Float32 => (ns,),
-                action=Float32 => (na,),
             ),
             sampler=BatchSampler{SS′ART}(
-                batch_size=128,
+                batch_size=2048,
```
HenriDeh, line 59: I lowered this batchsize to 128 in a previous PR because the NFQ tests were taking a lot of time. It used to be 10000. Can you check if going back up to 2048 significantly increases the run time ?

HenriDeh (commented):
Good changes, this implements NFQ in a more idiosyncratic way that reflects better RL.jl's design. 
I only have one question (see below).

HenriDeh:
> Unrelated, but during testing I tried to work with the EpisodeSampleRatioController and didn't manage to get it working: it complains that NamedTuple has no field terminal. I didn't find any examples of it in this repo or in RLTrajectories, so that may be worthwile to pursue in another issue.

Yes it's new and unused at the moment. Can you share a stacktrace, I'll open a PR to fix this.

src/ReinforcementLearningExperiments/deps/experiments/experiments/DQN/JuliaRL_NFQ_CartPole.jl
CasBex: On my CPU, the test with batch size 2048 takes 154 s and with batch size 128 takes 8.6 seconds. I modified the batch size to get better results, but notice now that even with batch size 128 they are quite good. I will put it back to 128

HenriDeh: great !

HenriDeh (approved):

PR #985, opened 2023-10-10
Fix and refactor SAC

### LINKED ISSUE #970: Fixing SAC Policy
I believe SAC is still an open issue, as the current implementation is unable to learn a viable policy. Even the JuliaRL_SAC_Pendulum [example](https://juliareinforcementlearning.org/docs/experiments/experiments/Policy%20Gradient/JuliaRL_SAC_Pendulum/#JuliaRL\\_SAC\\_Pendulum) does not show signs of improvement as the number of episodes grows. I've been reading the original [paper](https://arxiv.org/abs/1812.05905) for the past days and I believe I found the reason for this behaviour and the way to fix it.

The problem, I think, is not the SAC code itself, but the GaussianNetwork implementation (available in networks.jl in main branch, neural_network_approximator.jl in my version). Here is the [current](https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningCore/src/utils/networks.jl) implementation

```
function (model::GaussianNetwork)(rng::AbstractRNG, s; is_sampling::Bool=false, is_return_log_prob::Bool=false)
    x = model.pre(s)
    μ, raw_σ = model.μ(x), model.σ(x)
    σ = clamp.(raw_σ, model.min_σ, model.max_σ)
    if is_sampling
        z = ignore_derivatives() do
            noise = randn(rng, Float32, size(μ))
            μ .+ σ .* noise
        end
        if is_return_log_prob
            logp_π = sum(normlogpdf(μ, σ, z) .- (2.0f0 .* (log(2.0f0) .- z .- softplus.(-2.0f0 .* z))), dims=1)
            return z, logp_π
        else
            return z
        end
    else
        return μ, σ
    end
end
```
After carefully reading SAC details, I made a couple of changes to this code. Here is my version.
```
function (model::GaussianNetwork)(rng::AbstractRNG, s; is_sampling::Bool=false, is_return_log_prob::Bool=false)
    x = model.pre(s)
    μ, raw_logσ = model.μ(x), model.logσ(x)
    logσ = clamp.(raw_logσ, log(model.min_σ), log(model.max_σ))
    if is_sampling
        σ = exp.(logσ)
        noise = randn(rng, Float32, size(μ))
        z = μ .+ σ .* noise
	action = tanh.(z)
        if is_return_log_prob
	    ϵ = 1.0f-10        
	    logp_π = sum(normlogpdf(μ, σ, z) .- log.(1 .- action .^ 2 .+ ϵ), dims=1)
            return action, logp_π
        else
            return action
        end
    else
        return μ, logσ
    end
end
```
Here are the changes I made

1.  I used a squashing function, so the action output is action = tanh.(z), where z = μ .+ σ .* noise. See Appendix C. That means the environments needs to map the [-1,1] value to the real action, but that doesn't seem to be a major constraint.
2.  I changed logp_π, see equation 26. Added a small constant for stability. Honestly I have no idea where the other equation came from.
3.  Removed the ignore block. I am not sure about why, as I am not very family with Zygote, but results showed me this part is important. Possibly because it helps to implement something similar to equation 10

The σ = exp.(logσ) part is obsolete now, but that is not really important to the general idea. Normalizer field also seems to be unimportant.

I ran my own environment after these changes and results were very good, close to the results I got with DDPG. α, the temperature, was always positive (I noticed it was negative sometimes with the old implementation) and its value seems to decrease with the number of episodes. That shows the policy is becoming more deterministic and that the agent is learning. 

For a real implementation, μ last layer should not be relu, as it only outputs positive numbers and action will be always positive (unless noise is very negative). I simply used identity. 

PPO also employs Gaussian Networks. I am not familiar with its code so I am not sure how these changes would impact it. Nevertheless, I am confident that is the correct implementation for SAC. Feedback is welcome.

HenriDeh:
Following #970, I took it up on me to fix this (@gggoes you don't need to do it, but you can review the PR if you want). I did it because I'll actually need SAC but also because I think RL.jl is so obscure in it's internal workings that it's probably way too much work for someone who's not already into deep like me. 

Anyway. To fix SAC I had to implement a variant of the GaussianNetwork, I called it SoftGaussianNetwork but it's not a very good name because there's nothing soft about it (I'm open to alternative ideas). The difference between the two is that SoftGaussianNetwork is differentiable through the reparameterization trick (action sampling) and that it necessarily uses a tanh squash function. The latter could have been optional but I decided to stick with the algorithm described in the literature for now. A corollary of the discussion in #970 is that GaussianNetwork had an incorrect logpdf since I removed the tanh activation. It now accepts both identity and tanh and will have the correct output in both cases. 

The batch_size => batchsize changes are me being annoyed that we sometimes had one and sometimes the other, so I changed them all to batchsize. 

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?
 
closes #970.


src/ReinforcementLearningDatasets/README.md
```diff
@@ -21,7 +21,7 @@ julia> samples = Iterators.take(ds)
 Base.Iterators.Take{D4RLDataSet{Random.MersenneTwister}}(D4RLDataSet{Random.MersenneTwister}(Dict{Symbol, Any}(:reward => Float32[0.9236555, 0.8713692, 0.92237693, 0.9839225, 0.91540813, 0.8331875, 0.8102179, 0.78385466, 0.7304337, 0.6942671  …  5.0350657, 5.005931, 4.998442, 4.986662, 4.9730926, 4.9638906, 4.9503803, 4.9326644, 4.8952913, 4.8448896], :state => Float32[1.2521756 1.2519351 … 
 ...
 ```
-`ds` is of the type `D4RLDataset` which consists of the entire dataset along with some other information about the dataset. `samples` are in the form of `SARTS` with batch_size 256.
+`ds` is of the type `D4RLDataset` which consists of the entire dataset along with some other information about the dataset. `samples` are in the form of `SARTS` with batchsize 256.
```
jeremiahpslewis, line 24:
```suggestion
`ds` is of the type `D4RLDataset` which consists of the entire dataset along with some other information about the dataset. `samples` are in the form of `SARTS` with `batchsize` 256.
```

jeremiahpslewis:
LGTM. One note, I would bump the versions by a whole minor release, e.g. `0.14` -> `0.15`, as the `batchsize` renaming is definitely a breaking change.

jeremiahpslewis (changes requested):
See above

HenriDeh:
> LGTM. One note, I would bump the versions by a whole minor release, e.g. `0.14` -> `0.15`, as the `batchsize` renaming is definitely a breaking change.

Correct, there you go.

jeremiahpslewis (dismissed):

jeremiahpslewis (approved):

HenriDeh:
Sorry I keep finding compat entries that need updating before the tests can run. Maybe wait for the tests to be green before approving.

jeremiahpslewis (approved):

PR #995, opened 2023-10-20
Conservative Q-Learning

HenriDeh:
This PR introduces the CQL algorithm by reusing the existing SAC implementation.
This is an offline algorithm which means that it does not collect new data as the policy is trained. To this means, I had to introduced a new type of Agent, the OfflineAgent, that essentially acts identically to Agent except that it does not update! the trajectory. 

I made an experiment on Pendulum and it works pretty well, as well as SAC in fact. 

I also added the possibility to directly setup an online finetuning run an the PostExperimentStage, although this could be done manually by the user. I added it so that it could fit in a single `Experiment` for the tests.

PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


src/ReinforcementLearningCore/Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ReinforcementLearningCore"
 uuid = "de1b191a-4ae0-4afa-a27b-92d07f46b2d6"
-version = "0.15"
+version = "0.15.0"
```
jeremiahpslewis, line 3: Version bump to `15.1`?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -27,13 +27,31 @@ mutable struct Agent{P,T} <: AbstractPolicy
...
+        end
+        agent
+    end
+end
+
 Agent(;policy, trajectory) = Agent(policy, trajectory)
 
-RLBase.optimise!(agent::Agent, stage::S) where {S<:AbstractStage} = RLBase.optimise!(TrajectoryStyle(agent.trajectory), agent, stage)
-RLBase.optimise!(::SyncTrajectoryStyle, agent::Agent, stage::S) where {S<:AbstractStage} = RLBase.optimise!(agent.policy, stage, agent.trajectory)
+RLBase.optimise!(agent::Union{Agent,OfflineAgent}, stage::S) where {S<:AbstractStage} = RLBase.optimise!(TrajectoryStyle(agent.trajectory), agent, stage)
```
jeremiahpslewis, line 50: This seems awkward…can we not have `OfflineAgent <: Agent` so we don’t need the union type? I suppose even having a new abstract type `AbstractAgent` if necessary would be preferable to the union type, right?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -27,13 +27,31 @@ mutable struct Agent{P,T} <: AbstractPolicy
     end
 end
 
+mutable struct OfflineAgent{P,T,B,R} <: AbstractPolicy
+    policy::P
+    trajectory::T
+    behavior_agent::B #a behavior agent to fill the trajectory. Leave nothing if the trajectory is prefilled. Should share the same trajectory as the parent OfflineAgent
```
jeremiahpslewis, line 33: The `behavior_` prefixes are not immediately clear / easy to grok for me, not sure if there’s another way of naming this?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -27,13 +27,31 @@ mutable struct Agent{P,T} <: AbstractPolicy
     end
 end
 
+mutable struct OfflineAgent{P,T,B,R} <: AbstractPolicy
```
jeremiahpslewis, line 30: Can you add a doc string here?

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
jeremiahpslewis: Maybe we can have a `behavior::OfflineBehavior` attribute? Added advantage would be that `OfflineAgent` would not need to be mutable, because the mutable elements could be stored in `OfflineBehavior`. Note: pretty sure `OfflineBehavior` is not a great naming idea, but you probably have a better suggestion..

src/ReinforcementLearningCore/src/policies/agent/agent_base.jl
```diff
@@ -27,13 +27,31 @@ mutable struct Agent{P,T} <: AbstractPolicy
...
+    policy::P
+    trajectory::T
+    behavior_agent::B #a behavior agent to fill the trajectory. Leave nothing if the trajectory is prefilled. Should share the same trajectory as the parent OfflineAgent
+    behavior_steps::Int #steps to fill the trajectory (defaults to capacity of )
+    behavior_reset_condition::R #the reset condition of the environment.
+    function OfflineAgent(policy::P, trajectory::T, behavior_agent = nothing, behavior_steps = ReinforcementLearningTrajectories.capacity(trajectory.container.traces), behavior_reset_condition = ResetAtTerminal()) where {P<:AbstractPolicy, T<:Trajectory}
+        if behavior_steps == Inf
+            @error "behavior_steps is infinite, please provide a finite integer."
```
jeremiahpslewis, line 38:
```suggestion
            @error "`behavior_steps` is infinite, please provide a finite integer."
```

src/ReinforcementLearningZoo/src/algorithms/offline_rl/CQL_SAC.jl
```diff
@@ -0,0 +1,110 @@
...
+        action_sample_size::Int = 10,
+        α_cql::Float32 = 0f0,
+        α_lr::Float32 = 1f-3,
+        τ_cql::Float32 = 5f0,
+        α_cql_autotune::Bool = true,
+        cons_weight::Float32 = 1f0, #multiplies the Q difference before substracting tau, used to scale with the rewards.
+        finetune_experiment::E = nothing #Provide an second experiment to run at PostExperimentStage to finetune the sac policy, typically with an agent that uses the sac policy. Leave nothing if no finetuning is desired.
+    )
+
+    Implements the Conservative Q-Learning algorithm [1] in its continuous variant on top of the SAC algorithm [2]. CQLSACPolicy wraps a classic SACPolicy whose networks will be trained normally, except for the additional conservative loss.
```
jeremiahpslewis, line 16:
```suggestion
    Implements the Conservative Q-Learning algorithm [1] in its continuous variant on top of the SAC algorithm [2]. `CQLSACPolicy` wraps a classic `SACPolicy` whose networks will be trained normally, except for the additional conservative loss.
```

src/ReinforcementLearningZoo/src/algorithms/offline_rl/CQL_SAC.jl
```diff
@@ -0,0 +1,110 @@
...
+        α_cql::Float32 = 0f0,
+        α_lr::Float32 = 1f-3,
+        τ_cql::Float32 = 5f0,
+        α_cql_autotune::Bool = true,
+        cons_weight::Float32 = 1f0, #multiplies the Q difference before substracting tau, used to scale with the rewards.
+        finetune_experiment::E = nothing #Provide an second experiment to run at PostExperimentStage to finetune the sac policy, typically with an agent that uses the sac policy. Leave nothing if no finetuning is desired.
+    )
+
+    Implements the Conservative Q-Learning algorithm [1] in its continuous variant on top of the SAC algorithm [2]. `CQLSACPolicy` wraps a classic `SACPolicy` whose networks will be trained normally, except for the additional conservative loss.
+    CQLSACPolicy contains the additional hyperparameters that are specific to this method. α_cql is the lagrange penalty for the conservative_loss, it will be automatically tuned if ` α_cql_autotune = true`. `cons_weight` is a scaling parameter
```
jeremiahpslewis, line 17:
```suggestion
    `CQLSACPolicy` contains the additional hyperparameters that are specific to this method. α_cql is the lagrange penalty for the conservative_loss, it will be automatically tuned if ` α_cql_autotune = true`. `cons_weight` is a scaling parameter 
```

src/ReinforcementLearningZoo/src/algorithms/offline_rl/CQL_SAC.jl
```diff
@@ -0,0 +1,110 @@
...
+        finetune_experiment::E = nothing #Provide an second experiment to run at PostExperimentStage to finetune the sac policy, typically with an agent that uses the sac policy. Leave nothing if no finetuning is desired.
+    )
+
+    Implements the Conservative Q-Learning algorithm [1] in its continuous variant on top of the SAC algorithm [2]. `CQLSACPolicy` wraps a classic `SACPolicy` whose networks will be trained normally, except for the additional conservative loss.
+    CQLSACPolicy contains the additional hyperparameters that are specific to this method. α_cql is the lagrange penalty for the conservative_loss, it will be automatically tuned if ` α_cql_autotune = true`. `cons_weight` is a scaling parameter 
+    which may be necessary to decrease if the scale of the Q-values is large. `τ_cql` is the threshold of the lagrange conservative penalty.
+    See SACPolicy for all the other hyperparameters related to SAC.
+
+    If desired, you can provide an `Experiment(agent, env, stop_condition, hook)` to finetune_experiment to finish the training with a finetuning run. `agent` should be a normal `Agent` with policy being `sac`, an environment to finetune on.
```
jeremiahpslewis, line 21:
```suggestion
    If desired, you can provide an `Experiment(agent, env, stop_condition, hook)` to `finetune_experiment` to finish the training with a finetuning run. `agent` should be a normal `Agent` with policy being `sac`, an environment to finetune on. 
```

src/ReinforcementLearningZoo/src/algorithms/offline_rl/offline_rl.jl
```diff
@@ -7,3 +7,5 @@ include("FisherBRC.jl")
 include("PLAS.jl")
 include("ope/ope.jl")
 include("common.jl")
+=#
+include("CQL_SAC.jl")
```
jeremiahpslewis, line 11:
```suggestion
include("CQL_SAC.jl")

```

jeremiahpslewis (changes requested):
Just a few minor things, mainly documentation. Can you also maybe add a unit test for the agent?

HenriDeh:
Thanks for the review. I don't think OfflineBehavior is a terrible name, I couldn't figure out a better one (I'm not very imaginative when it comes to these things). 

I added the requested changes. Furthermore, I was still unsure about including fine-tuning directly in the policy, but I have since changed my mind. Fine-tuning is better thought of as running a second algorithm (SAC in this case) after the offline pretraining (CQL), rather than the penultimate step of CQL. Additionally, I think it simplifies the CQL struct to not have a nested subpolicy. OfflineAgent is already quite complicated with the OfflineBehavior subagent. Therefore, I'm removing it, unless you think it was better before?

src/ReinforcementLearningCore/src/policies/agent/offline_agent.jl
```diff
@@ -0,0 +1,76 @@
...
+
+Used to provide an OfflineAgent with a "behavior agent" that will generate the training data
+at the `PreExperimentStage`. If `agent` is `nothing` (by default), does nothing. The `trajectory` of agent should 
+be the same as that of the parent `OfflineAgent`.
+`steps` is the number of data elements to generate, defautls to the capacity of the trajectory.
+`reset_condition` is the episode reset condition for the data generation (defaults to `ResetAtTerminal()`).
+
+The behavior agent will interact with the main environment of the experiment to generate the data.
+"""
+struct OfflineBehavior{A <: Union{<:Agent, Nothing}, R}
```
jeremiahpslewis, line 14: Is `<:Agent` right here or should it be `AbstractAgent`?

jeremiahpslewis (changes requested):
This looks awesome now, one last comment above, after looking into the references, I get the `Offline` naming.

src/ReinforcementLearningCore/src/policies/agent/offline_agent.jl
HenriDeh: It must be a classic Agent, otherwise it won't generate data, which is the point.

jeremiahpslewis (approved):

PR #1040, opened 2024-03-11
Tabular Approximator fixes (pre v0.11 changes)

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh:
I see that you are still working on this, would you like a review now or later when it is ready ?

jeremiahpslewis:
@HenriDeh I think it's ready for a review; unless you insist, I will not fix the RLZoo issues as RLZoo is 'archived' for v0.11. If you want to review https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/pull/1041 first, then all remaining tests will pass. ;)

src/ReinforcementLearningCore/src/policies/learners/abstract_learner.jl
```diff
@@ -9,7 +9,7 @@ Base.show(io::IO, m::MIME"text/plain", L::AbstractLearner) = show(io, m, convert
 
 # Take Learner and Environment, get state, send to RLCore.forward(Learner, State)
 function forward(L::Le, env::E) where {Le <: AbstractLearner, E <: AbstractEnv}
-    env |> state |> Flux.gpu |> (x -> forward(L, x)) |> Flux.cpu
+    env |> state |> (x -> forward(L, x))
```
HenriDeh, line 12: This is minor, and not even yours, but the Julian style is to use lower case for variables and start with a capital for types. I think L::Le should become "learner".

src/ReinforcementLearningCore/src/policies/learners/target_network.jl
```diff
@@ -33,11 +33,32 @@ mutable struct TargetNetwork{M}
...
-function TargetNetwork(network; sync_freq = 1, ρ = 0f0)
+"""
+    TargetNetwork(network; sync_freq = 1, ρ = 0f0, use_gpu = false)
+
+Constructs a target network for reinforcement learning.
+
+# Arguments
+- `network`: The main network used for training.
+- `sync_freq`: The frequency (in number of calls to `optimise!`) at which the target network is synchronized with the main network. Default is 1.
+- `ρ`: The interpolation factor used for updating the target network. Must be in the range [0, 1]. Default is 0.
```
HenriDeh, line 44:
```suggestion
- `ρ`: The interpolation factor used for updating the target network. Must be in the range [0, 1]. Default is 0 (the old weights are completely replaced by the new ones).
```

HenriDeh:
Unless we change its scope, I think Approximators still belong in RLCore as they are common components in RL algorithms. The aliasing is a good idea. But presently it means that the state and action spaces must be [1... n]. While this is a common case, it would be nice to offer the possibility of arbitrary spaces. This also requires to know the state space beforehand. Sometimes this is not the case.
One possibility is to also make an alias for an Approximator with a Dict instead of an Array where new states/station-actions are added to the keys when they are encountered for the first time. I don't know what you think.

jeremiahpslewis:
@HenriDeh I'm not sure I understand you entirely, it's always possible to create arbitrary further (aliased) approximators, that wouldn't be a breaking change; an approximation with a model type = Dict is already workable (afaict). Do you see code in RLCore which would cause an `Approx(model=Dict())` to break?

HenriDeh:
For example, one classic problem used in Sutton and Barton is a 2-d gridworld (https://towardsdatascience.com/training-an-agent-to-beat-grid-world-fac8a48109a8). The action-space is {up, down, left, right}, and the state-space is [1...n]^2. Of course, in this case it's easy to biject each to an integer, but not in general. 
But say I want to do a Q-value `Approximator` with a Dict, then it will not use the same api as a `TabularQApproximator`. So a tabular Q-based algorithm will use `forward(A, s, a)`, for example, whereas the Dict Approximator will use `forward(A, (s,a)`.

jeremiahpslewis:
Ok, so we would then have something like `DictApproximator = Approximator{m} where {m <: Dict}` and could dispatch on this type signature? As long as we can get more specific in terms of type, we should be able to dispatch differently on this (perhaps by splatting the tuple before passing it along).

HenriDeh (approved):

jeremiahpslewis:
@HenriDeh would you prefer we change the naming to `ArrayApproximator`? Is that more transparent?

HenriDeh:
> Ok, so we would then have something like DictApproximator = Approximator{m} where {m <: Dict} and could dispatch on this type signature? As long as we can get more specific in terms of type, we should be able to dispatch differently on this (perhaps by splatting the tuple before passing it along). 
And anything that is using forward(approximator, env) can be handled likewise (right?)

Yes. This is not urgent so let's move on because the priority is to get a working v0.11. 
Later we could consider another approach: instead of a general `Approximator`, have a `VApproximator{A}`, a `QAprroximator{A}`, and even a `PolicyApproximator{A}` where `A` is the type of the approximator. Because really these three are the things that we approximate in RL and they have distinct APIs. This should be doable without trouble.

 > @HenriDeh would you prefer we change the naming to ArrayApproximator? Is that more transparent?

I prefer the current naming because it explicitly says what function is approximated.

PR #1041, opened 2024-03-11
Swap RLZoo for RLFarm in CI, drop RLExperiments

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,8 +96,8 @@ jobs:
             Pkg.test("ReinforcementLearningCore", coverage=true)'
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v3
-  test-RLZoo:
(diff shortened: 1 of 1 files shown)


HenriDeh (approved):

HenriDeh:
Experiments used to be the de facto place where algorithms were CI tested. Will we expect this to be done in RLFarm going forward?

jeremiahpslewis:
I think that makes sense, yes. And with explicit outcome/correctness checks which go beyond whether the code runs through?

HenriDeh:
My take then is that these tests should not be like those we had in RLExperiments. They should only test that internal functions (losses for example) do what they are expected to do, and that the training loop works. Testing the performance of the algorithms (both in terms of speed and learning) could be done in the future ad hoc RLExperiments.

jeremiahpslewis:
Sounds good!

PR #1046, opened 2024-03-13
Fix reset and stop conditions

jeremiahpslewis:
This pull request updates the reset and stop conditions in the code to use a consistent `check!` in-place interface. It is a step on the way to getting a high level of test coverage


src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -58,12 +58,12 @@ function _stop_after_step(s::StopAfterStep)
     res
 end
 
-function check_stop(s::StopAfterStep, args...)
+function check!(s::StopAfterStep, args...)
```
HenriDeh, line 61: Since we are renaming the conditions, should this be renamed to `StopAfterNSteps` to match the naming of the reset condition ?

src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -91,15 +91,15 @@ function StopAfterEpisode(episode; cur = 0, is_show_progress = true)
     StopAfterEpisode(episode, cur, progress)
 end
 
-function check_stop(s::StopAfterEpisode{Nothing}, agent, env)
+function check!(s::StopAfterEpisode{Nothing}, agent, env)
```
HenriDeh, line 94: Ibid, `StopAfterNEpisodes`

src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -167,7 +167,7 @@ Return `true` if the environment is terminated.
 """
 struct StopWhenDone <: AbstractStopCondition end
 
-check_stop(s::StopWhenDone, agent, env) = is_terminated(env)
+check!(s::StopWhenDone, agent, env) = is_terminated(env)
```
HenriDeh, line 170: `StopIfEnvTerminated`, since it does exactly the same check as the reset condition.

HenriDeh (approved):

PR #1048, opened 2024-03-14
Drop Functors and use Flux.@layer

jeremiahpslewis:
This pull request drops the use of Functors and replaces it with Flux.@layer, fixing a Flux.jl warning.


src/ReinforcementLearningCore/src/policies/learners/target_network.jl
```diff
@@ -61,9 +60,7 @@ function TargetNetwork(network::Approximator; sync_freq = 1, ρ = 0f0, use_gpu =
     return TargetNetwork(network, target, sync_freq, ρ, 0)
 end
 
-@functor TargetNetwork (network, target)
-
-Flux.trainable(model::TargetNetwork) = (model.network,)
+Flux.@layer TargetNetwork trainable=(network,)
```
HenriDeh, line 63: Weird that target was trainable before. I wonder if this used to cause bugs.

HenriDeh (approved):

PR #1049, opened 2024-03-15
Fix naming consistency and add missing hook tests

jeremiahpslewis:
This pull request includes changes to make the naming consistent for the `DoEveryNSteps` hook and adds missing tests for hooks.


src/ReinforcementLearningCore/src/core/hooks.jl
```diff
@@ -286,26 +286,26 @@ function Base.push!(hook::TimePerStep, ::PostActStage, agent, env)
...
     n::T
     t::T
 
-    function DoEveryNStep(f; n=1, t=0)
+    function DoEveryNSteps(f; n=1, t=0)
         new{typeof(f),Int64}(f, n, t)
     end
 
-    function DoEveryNStep{T}(f; n=1, t=0) where {T<:Integer}
+    function DoEveryNSteps{T}(f; n=1, t=0) where {T<:Integer}
```
HenriDeh, line 303: Couldn't we have 
```
mutable struct DoEveryNSteps{F} <: AbstractHook where {F}
    f::F
    n::Int
    t::Int
end
```
instead of supporting multiple Int precisions ? This would make the last constructor unnecessary. I feel like this is extra code to compile that noone will ever need and which may lead to sneaky bugs.

HenriDeh (approved):

PR #1052, opened 2024-03-20
Epsilon Speedy Explorer

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?

### DIFF
--- a/src/ReinforcementLearningFarm/src/algorithms/explorers/epsilon_speedy_explorer.jl
+++ b/src/ReinforcementLearningFarm/src/algorithms/explorers/epsilon_speedy_explorer.jl
@@ -6,6 +6,7 @@ import ReinforcementLearningBase: RLBase
 using FillArrays: Trues
 using Random
 using Distributions: Categorical
+using Base
 
 # Epsilon Greedy Explorer for AIAPC Zoo
 # Note: get_ϵ function in RLCore takes: 600.045 ns (6 allocations: 192 bytes)
@@ -16,7 +17,7 @@ using Distributions: Categorical
 struct EpsilonSpeedyExplorer{R} <: AbstractExplorer
     β::Float64
     β_neg::Float64
-    step::Ref{Int}
+    step::Base.RefValue{Int}
     rng::R
 end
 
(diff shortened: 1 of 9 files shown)


src/ReinforcementLearningFarm/src/algorithms/explorers/epsilon_speedy_explorer.jl
```diff
@@ -0,0 +1,92 @@
...
+
+`EpsilonSpeedyExplorer` is an explorer that selects the action with the maximum value with probability `1 - ϵ` and selects a random action with probability `ϵ`.
+The probability of selecting a random action is given by `exp(β * -t)`, where `t` is the number of times `plan!` has been called.
+`EpsilonSpeedyExplorer` differs from `EpsilonGreedyExplorer` in that it uses the `exp` function to calculate the probability of selecting a random action over the full range of `t` and only accepts one argument, `β`.
+
+"""
+struct EpsilonSpeedyExplorer{R} <: AbstractExplorer
+    β::Float64
+    β_neg::Float64
+    step::Base.RefValue{Int}
```
HenriDeh, line 22: Is it really faster to use a Ref instead of a mutable struct with ::Int ?

src/ReinforcementLearningFarm/src/algorithms/explorers/epsilon_speedy_explorer.jl
```diff
@@ -0,0 +1,92 @@
...
+    else
+        ϵ / n
+    end
+end
+
+function RLBase.prob(s::EpsilonSpeedyExplorer, values::A, mask::M) where {I<:Real, A<:AbstractArray{I}, M<:Union{BitVector, Vector{Bool}}}
+    ϵ, n = get_ϵ(s), length(values)
+    probs = zeros(n)
+    probs[mask] .= ϵ / sum(mask)
+    probs[RLCore.findmax_masked(values, mask)[2]] += 1 - ϵ
```
HenriDeh, line 90: Unless I missed it, there should be a test to check that this returns the expected distribution for the actions.

src/ReinforcementLearningFarm/src/algorithms/explorers/epsilon_speedy_explorer.jl
jeremiahpslewis: good question. timings are identical:

```
# 21.079 ns (<0.01 allocs: 0.130 bytes)
# 22.028 ns (<0.01 allocs: 0.136 bytes)
```

When there's only one struct element that needs to be mutable and it's a float, I like to use a `Ref` to eliminate accidental mutations, happy to change this...

HenriDeh (dismissed):

PR #1054, opened 2024-03-21
Fix abstract_learner for multiplayer games

jeremiahpslewis:
This pull request fixes the abstract_learner for multiplayer games. It adds a new function `forward(learner::L, env::E, player::Symbol)` that takes a learner, environment, and player symbol as arguments and sends the state to `RLCore.forward(Learner, State)`. This ensures that the learner can handle multiplayer games correctly.

It also cleans up a few other minor issues along the way.

### DIFF
--- a/src/ReinforcementLearningCore/src/core/stop_conditions.jl
+++ b/src/ReinforcementLearningCore/src/core/stop_conditions.jl
@@ -18,13 +18,13 @@ The result of `stop_conditions` is reduced by `reducer`. The default `reducer` i
 struct ComposedStopCondition{S,T} <: AbstractStopCondition
     stop_conditions::S
     reducer::T
-    function ComposedStopCondition(stop_conditions...; reducer = any)
-        new{typeof(stop_conditions),typeof(reducer)}(stop_conditions, reducer)
+    function ComposedStopCondition(stop_conditions...; reducer::T = any)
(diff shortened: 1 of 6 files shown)


src/ReinforcementLearningCore/src/core/stop_conditions.jl
```diff
@@ -15,16 +15,16 @@ abstract type AbstractStopCondition end
 
 The result of `stop_conditions` is reduced by `reducer`. The default `reducer` is the `any` function, which means that the condition is true when any one of the `stop_conditions...` is true. Can be replaced by any function returning a boolean. For example `reducer = x->sum(x) >= 2` will require at least two of the conditions to be true.
 """
-struct ComposedStopCondition{S,T} <: AbstractStopCondition
+struct ComposedStopCondition{S,reducer} <: AbstractStopCondition
     stop_conditions::S
-    reducer::T
+    reducer
```
HenriDeh, line 20: I had no idea you could do that!

HenriDeh (approved):

src/ReinforcementLearningCore/src/core/stop_conditions.jl
jeremiahpslewis: me neither! ...but I was getting desperate 😆

PR #1062, opened 2024-03-26
Add note about RLZoo deprecation

jeremiahpslewis:
PR Checklist

- [ ] Update NEWS.md?
- [ ] PR Review?


HenriDeh (commented):
Maybe we can add an @warn in the `__init__` method of the package. This way users will be warned when loading it.

jeremiahpslewis:
@HenriDeh What about an @info when loading RL.jl?

HenriDeh:
That also works.

Issue #1001, opened 2023-11-14
TicTacToeEnv allows illegal moves


colintbowers:
Apologies in advance if this is deliberate behavior, but it seemed odd to me that `TicTacToeEnv` allows illegal moves. For example:
```
env = TicTacToeEnv()
env(1)
env(1)
```
allows o and x to go in the same square (top-left). Note that a call to:
```
is_terminated(env)
```
will now error with the message:
```
ERROR: KeyError: key TicTacToeEnv([0 1 1; 1 1 1; 1 1 1;;; 1 0 0; 0 0 0; 0 0 0;;; 1 0 0; 0 0 0; 0 0 0], ReinforcementLearningEnvironments.Cross()) not found
```
An implication of this behavior is:
```
env = TicTacToeEnv()
env(1)
env(2)
env(1)
```
will now place the board in a state where there is one o and one x, but it is o's turn.

I'm brand new to this package, but it seems to me that the fix should just be to add a check to the function:
```
function (env::TicTacToeEnv)(action::CartesianIndex{2})
    env.board[action, 1] = false
    env.board[action, Base.to_index(env, env.player)] = true
    env.player = !env.player
end
```
maybe something like `!env.board[action,1] && error("some message")``, since if that square is already false, it means a move has already been played there. Or perhaps there is some other desired return when an illegal move is played. I'm not very familiar with this package (or topic) yet.

Cheers,

Colin

PR #1094, opened 2025-06-07
Fix for TicTacToeEnv allows illegal moves #1001

SanteriVtj:
TicTacToeEnv allows one state to be played multiple times by the same agent. This condition should prevent this behavior and throw an error in case the state is played multiple times.


jeremiahpslewis (approved):

chrbauer:
Question: should act! not also call next_player? Or how can i use this env? 
The example Pluto notebook is quite broken. (e.g no act! is used)

https://juliareinforcementlearning.org/ReinforcementLearningAnIntroduction.jl/notebooks/Chapter01_Tic_Tac_Toe.html

jeremiahpslewis:
The test should be an ok example: https://github.com/JuliaReinforcementLearning/ReinforcementLearning.jl/blob/main/src/ReinforcementLearningEnvironments/test/environments/examples/tic_tac_toe.jl

